PR #8, opened 2019-07-04
Improve/simplify C++ book recommandations

jonatack:
Remove "C++ Primer Plus". Here are 3 links that explain why:

  - https://www.reddit.com/r/cpp/comments/13g6s1/which_c_primer_book_is_better/
  - https://www.quora.com/Between-C++-Primer-Plus-and-C++-Primer-which-is-better-and-why
  - http://www.cplusplus.com/forum/beginner/106097/

[C++ Primer, 5th Edition](https://www.amazon.com/Primer-5th-Stanley-B-Lippman/dp/0321714113/) and the dated but compact and well-regarded [Accelerated C++](https://www.amazon.com/Accelerated-C-Practical-Programming-Example/dp/020170353X/) are introductory books I've worked through and can recommend. That said, individual readers' needs vary, and it may be best to simply provide a link to this (quite good) list of books organised by category:

  - https://stackoverflow.com/questions/388242/the-definitive-c-book-guide-and-list


jnewbery:
Much better. Thanks Jon!

PR #11, opened 2019-07-04
Doc: Fix command to launch Jekyll server

jonatack:
per usual practice and https://jekyllrb.com/docs

Omitting `bundle exec` can cause errors like:

```
You have already activated liquid 4.0.3, but your Gemfile requires liquid 4.0.0.
Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```

Cheers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ You'll need [Ruby & Jekyll](https://jekyllrb.com/docs/installation/) to run the 
 
 * Clone the repository and go into the directory
 * Run `bundle install`
-* Run `jekyll serve`
+* Run `bundle exec jekyll serve`
 * Go to http://localhost:4000
 
 ## Making a Post


jonatack:
Here's more info on `bundle exec`:

- https://bundler.io/v2.0/man/bundle-exec.1.html
- https://www.learnhowtoprogram.com/ruby/databases-with-active-record/bundle-exec

It avoids version conflicts by ensuring the bundle command operates in the current bundle environment. Sometimes Bundler can work without it, e.g. if only the system Ruby is installed or only one version of the gem exists, but it often won't, e.g. when multiple rubies or gem versions exist or if the user manages rubies with a version manager like rbenv or rvm.

jnewbery:
Will merge once #12 (or alternative) is merged

PR #16, opened 2019-07-23
doc: update README intro and add website URL

jonatack:
Trivial update to the README to:

1. update/simplify the introductory description because the website is no longer only about upcoming meetings; it now contains a history and log of past meetings, and

2. add the website URL to enable easily navigating from the repository to the website (something I've been missing, so this is the real reason).

Click here here to test this change:

https://github.com/jonatack/bitcoin-core-review-club.github.io/tree/add-website-url-to-readme-doc

Cheers!


jnewbery:
Looks good. Thanks!

PR #26, opened 2019-09-26
n16202: add p2p encryption core dev transcript

jonatack:
and (perhaps controversially) update an imprecise answer of mine to not send people looking in the wrong place.


jnewbery:
Good idea to add the link to the transcript, but I don't think we should be changing the transcripts after the fact. Can you remove that part of the commit?

jnewbery:
Feel free to add a **corrections** section at the end.

jonatack:
Good idea @jnewbery, thanks!

jnewbery:
Looks great. Thanks for your diligence 😁

PR #29, opened 2019-09-29
Improve meeting links and headers

jonatack:
I'm not sure if it's necessary to toggle the links between bold and regular style, as it changes the widths, but I followed the intention of the commit it builds on and hopefully fixes.

Edit: I separated the first commit into two, one for the fixes and one for purely content changes. Commits 2 and 3 are (quite optional) content-only edits.

### DIFF
--- a/_data/settings.yml
+++ b/_data/settings.yml
@@ -7,7 +7,7 @@ author:
 menu:
 # Uncomment to activate the menu looping function in the header.html file under _includes
 - {name: 'Home', url: ''}
-- {name: 'Previous meetings', url: 'previous_meetings.html'}
+- {name: 'Meetings', url: 'meetings.html'}
 
 social:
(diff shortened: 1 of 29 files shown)


jonatack:
FWIW I wanted to use Ruby inside these templates, like with ERB (Embedded Ruby) templates to abstract some repeated code out more functionally, but at first try I didn't figure out how without changing file types.

jnewbery:
I've taken these commits into #28. Thanks @jonatack !

> I wanted to use Ruby inside these templates, like with ERB (Embedded Ruby) templates to abstract some repeated code out more functionally

Yes, it's a shame there's a lot of repeated code across the different indexes. Perhaps we can use some clever Jekyll includes to remove that duplication, but I'm not an expert.

jonatack:
Same, looked at the Liquid docs for the first time today.

PR #31, opened 2019-09-30
Fixups and improvements

jonatack:
Builds on #28, feel free to pick and choose.

- Add missing components in the PR layout.

- Fix and format the HTML after I began to notice dangling/missing tags and mis-indented markup. This should also help to see future errors. I know this kind of change is annoying to review :(  

- Remove redundant PR links in the posts, since these links are now included in the PR layout. Before this change, many posts display the PR links twice.

- Ensure anchor tags (`#`) function correctly in the `meetings-components` and `meetings-hosts` listings when the target is near the bottom of the listing by adding bottom padding to the page.

### DIFF
--- a/_data/settings.yml
+++ b/_data/settings.yml
@@ -7,7 +7,7 @@ author:
 menu:
 # Uncomment to activate the menu looping function in the header.html file under _includes
 - {name: 'Home', url: ''}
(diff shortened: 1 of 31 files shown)


jonatack:
Closing; commits here were merged into #28.

PR #33, opened 2019-09-30
Remove trailing slashes from links to meeting pages

jonatack:
as a follow-up patch to #28 for 404s seen only in production.


jonatack:
Ugh, there is so much duplication instead of a single point of truth for these link-to's.

jonatack:
This should be ready now.

jnewbery:
Alternative at #34

jnewbery:
#34 seems to have fixed this. Thanks for your help!

PR #45, opened 2019-10-12
Add test setup and test all site links

jonatack:
TL|DR:
```
$ bundle update && gem install curb
$ make preview (if running the links test on the local server) 
$ rake
```
This PR sets up testing in a minimal way using Minitest, a simple, fast, tiny mini-framework in pure Ruby with no magic or DSL. It also adds a pretty test reporter because it's much nicer and a tiny dependency.

This allows running the tests from the command line with `rake test` or simply `rake`. Individual test files or cases can be run as well as per the comments in the Rakefile. There are integrations available for various editors, including vim and emacs, to run this from within the editor environment. 

--------
NEW: Added a test to check all the links in the site -- either locally after running `make preview` (default), or on the website in production, and display a report in the terminal.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -88,6 +88,3 @@ DEPENDENCIES
   minitest
   minitest-reporters
   rake
-
-BUNDLED WITH
-   2.1.4
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,8 +3,16 @@ GEM
   specs:
     addressable (2.7.0)
       public_suffix (>= 2.0.2, < 5.0)
+    ansi (1.5.0)
+    builder (3.2.4)
+    color_pound_spec_reporter (0.0.9)
+      bundler
+      minitest
+      minitest-reporters (>= 0.14.24)
+      rake
     colorator (1.1.0)
     concurrent-ruby (1.1.5)
+    curb (0.9.10)
     em-websocket (0.5.1)
       eventmachine (>= 0.12.9)
       http_parser.rb (~> 0.6.0)
@@ -44,6 +52,12 @@ GEM
       rb-fsevent (~> 0.10, >= 0.10.3)
       rb-inotify (~> 0.9, >= 0.9.10)
     mercenary (0.3.6)
+    minitest (5.14.0)
+    minitest-reporters (1.4.2)
+      ansi
+      builder
+      minitest (>= 5.0)
+      ruby-progressbar
     pathutil (0.16.2)
       forwardable-extended (~> 2.6)
     public_suffix (4.0.1)
@@ -52,6 +66,7 @@ GEM
     rb-inotify (0.10.0)
       ffi (~> 1.0)
     rouge (3.11.1)
+    ruby-progressbar (1.10.1)
     safe_yaml (1.0.5)
     sass (3.7.4)
       sass-listen (~> 4.0.0)
@@ -63,9 +78,16 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
+  color_pound_spec_reporter
+  curb
   jekyll (~> 3.8)
   jekyll-feed (~> 0.12.1)
   jekyll-paginate (~> 1.1)
   jekyll-seo-tag (~> 2.6)
   jekyll-sitemap (~> 1.3)
+  minitest
+  minitest-reporters
   rake
+
+BUNDLED WITH
+   2.1.4
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,22 @@ require 'date'
 require 'json'
 require 'net/http'
 require 'optparse'
+require 'rake/testtask'
+
+# To run all tests:
+#   rake (or) rake test
+#
+# To run one test file:
+#   rake test TEST=test/FILENAME
+#
+# To run an individual test in a test file:
+#   rake test TEST=test/FILENAME TESTOPTS=--name=TEST_NAME
+#
+desc 'Run all tests with `rake` or `rake test`'
+task default: :test # Make test runner the default rake task.
+Rake::TestTask.new do |task|
+  task.pattern = 'test/test_*.rb'
+end
 
 # These correspond to the GitHub labels used by Bitcoin Core.
 DESIRED_COMPONENTS = [
--- a/Gemfile
+++ b/Gemfile
@@ -9,5 +9,13 @@ gem "jekyll-sitemap", "~> 1.3"
 gem "jekyll-feed", "~> 0.12.1"
 gem "jekyll-seo-tag", "~> 2.6"
 
-# Ruby Make
+# Ruby make
 gem 'rake'
+
+# Testing
+gem 'minitest'
+gem 'minitest-reporters'
+gem 'color_pound_spec_reporter'
+
+# HTTP bindings to libcurl - https://github.com/taf2/curb
+gem 'curb'
--- a/README.md
+++ b/README.md
@@ -28,10 +28,8 @@ To run one test file: `rake TEST=test/test_rake_posts_new`
 To run an individual test in a test file:
 `rake TEST=test/test_rake_posts_new TESTOPTS=--name=test_rake_posts_new`
 
-Before running tests for the first time, you may need to run:
-```
-$ bundle update && gem install curb
-```
+Before running tests for the first time, you may need to run `bundle update`.
+
 Before running the `test_all_links` test (or all of the tests, which includes
 it), the site needs to be started up locally with `make preview`, or with `make
 clean preview` if any meeting logs were changed, to pre-render the logs through
--- a/README.md
+++ b/README.md
@@ -19,6 +19,23 @@ See the [CONTRIBUTING.md](CONTRIBUTING.md) doc for how to create a post for an u
 
 All site configurations are either contained in `_config.yml` or `_data/settings.yml`. Some data is duplicated between the two due to the way Jekyll injects variables, so be sure to update both.
 
+## Running the Test Suite
+
+To run all tests: `rake test` or just `rake`
+
+To run one test file: `rake TEST=test/test_rake_posts_new`
+
+To run an individual test in a test file:
+`rake TEST=test/test_rake_posts_new TESTOPTS=--name=test_rake_posts_new`
+
+Before running tests for the first time, you may need to run:
+```
+$ bundle update && gem install curb
+```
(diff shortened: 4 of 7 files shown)


jonatack:
```
$ rake
Started with run options --seed 872

TestRakePostsNew
 PASS (0.00s) :: test_rake_posts_new

Finished in 0.00065s
1 tests, 1 assertions, 0 failures, 0 errors, 0 skips
```

jonatack:
FWIW the codebase of a bundled gem e.g. in the Gemfile can be inspected with `bundle open <GEM_NAME>`

For example, `bundle open minitest`

Rakefile
```diff
@@ -4,6 +4,26 @@ require 'date'
 require 'json'
 require 'net/http'
 require 'optparse'
+require "rake/testtask"
+
+# To run all tests:
+#   rake (or) rake test
+#
+# To run one test file:
+#   ruby -Ilib:test test/FILENAME.rb
```
jnewbery: Remove this. Better to have just one way to do things.

Rakefile
```diff
@@ -4,6 +4,26 @@ require 'date'
...
+#   (or)
+#   rake test TEST=test/FILENAME.rb
+#
+# To run an individual test in a test file:
+#   ruby -Ilib:test test/FILENAME.rb -n TEST_NAME
```
jnewbery: Same as above.

Rakefile
```diff
@@ -4,6 +4,26 @@ require 'date'
...
+#   (or)
+#   rake test TEST=test/FILENAME.rb TESTOPTS=--name=TEST_NAME
+#
+desc 'Run all tests with `rake` or `rake test`'
+task default: :test # Make test runner the default rake task.
+Rake::TestTask.new do |task|
+  task.pattern = 'test/test_*.rb'
```
jnewbery: I don't understand ruby. Does this not cause `test_helper.rb` to run?

Rakefile
jonatack: Done.

Rakefile
jonatack: Done.

Rakefile
jonatack: Good eye! Yes, if the test files are not specified, e.g. `rake test` works, but `rake TEST=test/test_rake_posts_new.rb` will fail with `uninitialized constant Minitest` if the test file does not require `test_helper`.

FWIW we can name test files as '*_test.rb' if you prefer. In this case the test helper is never auto-run.

jonatack:
> Can you expand a bit more on what you expect the tests to do? Is it going to hit the github API for every existing post (which I think would cause the test to be very slow), or is it just checking that every post has all the correct page variables?

For the first iteration, I was thinking the tests could address these cases we were manually testing for one valid PR (plus one test case with a non-existent PR):

```
    no PR argument (fails)
    non-numeric PR argument (fails)
    numeric PR argument that doesn't match a PR (fails with HTTP 404)
    no host argument (fails)
    invalid date argument (fails)
    duplicate PR (fails)
    valid PR and host arguments, no date argument (succeeds)
    valid PR host and date arguments (succeeds with HTTP 200)
```
which would imply 2, maybe 3, API calls: minimum one HTTP 200, and one 404. The other cases should fail before the HTTP call. 

We could mock this response later on if we want to avoid making the 2 API calls, but then we lose detection if GH changes its REST API response. 

Performance seems initially less a constraint than the GitHub API rate limit: 60 REST API calls per hour, so could run the tests 30 times per hour. I could live with that initially, but... running a future `rake posts:update_all` command that calls every PR will hit the limit quickly. If we switch to GH's GraphQL API, the limit rises to 5000 per hour, we can request only the data fields we need, and even all the needed PRs in one call, but we have to pass an OAuth token which would add an `--auth $TOKEN` argument to pass.

jnewbery:
Oh, you're planning to add test cases for the `rake posts:new` script? That seems a bit like overkill to me. I thought this was more for testing that the data in all the posts is correct, there are no bad links, etc.

jonatack:
That could be done too using this test setup. For the moment I'm using it locally with tests written for the `rake posts:new` task to avoid having to test a bunch of edge cases manually when changing it. These will be helpful as well when doing the refactoring and `update` tasks I mention in the description at the top. Also, I noticed a couple more edge cases that could be fixed, and testing it with the automated tests by running `rake test` is nicer.

jonatack:
> testing that the data in all the posts is correct, there are no bad links, etc.

Added a test to check all the links and output the results, fixed a bad displayed link the test found in the PR layout, and updated the PR description. I think most of the 301s will be fixed by merging #158.

Gemfile.lock
```diff
@@ -63,9 +78,16 @@ PLATFORMS
...
   jekyll (~> 3.8)
   jekyll-feed (~> 0.12.1)
   jekyll-paginate (~> 1.1)
   jekyll-seo-tag (~> 2.6)
   jekyll-sitemap (~> 1.3)
+  minitest
+  minitest-reporters
   rake
+
+BUNDLED WITH
```
jnewbery: Checking out this branch and running `bundle install` gives me an error:

```
Traceback (most recent call last):
	4: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `<main>'
	3: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `eval'
	2: from /home/john/.rvm/gems/ruby-2.5.1/bin/bundle:23:in `<main>'
	1: from /home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:308:in `activate_bin_path'
/home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
```

Removing the `BUNDLED WITH` lines and running `bundle install` works.

Should we just remove these lines? I'm not familiar enough with ruby/bundler to understand what these things are doing.

jonatack:
@jnewbery apologies, just realized that I changed the test to `localhost:4000` by default instead of the production website, without updating the instructions at the top. Running the links test on localhost requires running `make preview` before launching the test. Do you have a preference for the default?

Gemfile.lock
jonatack: Could you run `bundle update` and see if that works?

jnewbery:
> Do you have a preference for running it by default on localhost or on the production website?

localhost I think.

Gemfile.lock
jonatack: You may also need to run `gem install curb`. Will update the PR description and maybe add some docs to the readme or contributing page.

jonatack: Added this info to README documentation about running the tests.

jonatack:
Rebased on master.

Gemfile.lock
jnewbery: Tried

```
 bundle update
Traceback (most recent call last):
	4: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `<main>'
	3: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `eval'
	2: from /home/john/.rvm/gems/ruby-2.5.1/bin/bundle:23:in `<main>'
	1: from /home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:308:in `activate_bin_path'
/home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
```

and after running `gem install curb`:

```
[john] /home/john/Code/docs/r2
→ gem install curb
Building native extensions. This could take a while...
Successfully installed curb-0.9.10
Parsing documentation for curb-0.9.10
Done installing documentation for curb after 0 seconds
1 gem installed
[john] /home/john/Code/docs/r2
→ bundle install
Traceback (most recent call last):
	4: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `<main>'
	3: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `eval'
	2: from /home/john/.rvm/gems/ruby-2.5.1/bin/bundle:23:in `<main>'
	1: from /home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:308:in `activate_bin_path'
/home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
[john] /home/john/Code/docs/r2
→ bundle update
Traceback (most recent call last):
	4: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `<main>'
	3: from /home/john/.rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:24:in `eval'
	2: from /home/john/.rvm/gems/ruby-2.5.1/bin/bundle:23:in `<main>'
	1: from /home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:308:in `activate_bin_path'
/home/john/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
```

Why do we need to add this `BUNDLED WITH` line?

jonatack: Bundler adds that line automatically (I didn't add it). What do you get when running `bundle -v`?

jonatack: I don't think it's an issue to remove it, but it's a symptom of an underlying issue.

jonatack:
Thanks for reviewing, @jnewbery!

Gemfile.lock
jonatack: Here's where the `BUNDLE WITH` line came from: it's autogenerated when the file was created by bundler version 2.x instead of 1.x. Explanations and how to upgrade (tldr: run `bundle update --bundler` to upgrade): https://bundler.io/guides/bundler_2_upgrade.html

PR #77, opened 2019-11-28
Add undesired labels to `rake posts:new` task

jonatack:
Added 'feature' and 'refactor' to the undesired labels array, after noticing while adding n16702 that these were ending up in the components array. Any others to add?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,8 +6,14 @@ require 'net/http'
 require 'optparse'
 
 UNDESIRED_LABELS = [
-  'refactoring', 'needs conceptual review', 'needs rebase', 'waiting for author'
+  'feature',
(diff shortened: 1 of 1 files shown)


jnewbery:
Could add 'brainstorming', 'bug', 'good first issue', 'hacktoberfest', 'needs backport', 'needs gitian build', 'needs release notes', 'questions and help', 'review club', 'review op'.

jnewbery:
I wonder if it's perhaps better to have DESIRED labels instead of UNDESIRED (since the list of components we're interested in probably won't change very much, but maintainers can add labels at any time)

jonatack:
Good idea -- proposed in #80.

PR #80, opened 2019-11-29
Filter components by desired for `rake posts:new`

jonatack:
rather than by undesired components currently as per discussion in PR #77.

LMK if any components used in the current repo tags are missing.


jnewbery:
Looks good. Thanks

PR #84, opened 2019-12-04
doc: add "After the Meeting" to CONTRIBUTING.md

jonatack:
and mention to hosts to open a PR after the session to add the meeting log.


jnewbery:
I don't think we should expect the guest host to do this. I'm happy to do it myself for all meetings. Probably far quicker for me to do it than the host.

We could also automate posting of the logs, but it's very little work for me.

jonatack:
All good. Closing.

PR #79, opened 2019-11-29
Fix vertical alignment of PRs

jnewbery:
Close #27

PR #86, opened 2019-12-05
Meetings pages improvements

jonatack:
These commits are suggestions that build on #79 with the following steps:

- Noticed that the new date table columns were being wrapped in narrow browser window sizes
- Added CSS to fix that, but it was not being applied
- This was due to extraneous HTML span tags inside the table markup. I removed those.
- To make sure I wasn't breaking anything, I began indenting the HTML more clearly.
- The indentation revealed a few issues like missing closing tags, so I fixed those.
- Added a `<br>` tag to the bottom of each meeting page to add a line of space so the text at the bottom of the page doesn't hug the window border as tightly; though it should probably be done with CSS instead
- The dates are no longer wrapped (yay) but that adds some padding to the left of the ">" separator, so I added CSS to add a bit of right padding to balance that out
- I'm not a front end expert and am unsure if pixels (`px`) are the best units to use for wide compatibility, noticing that the other CSS mostly uses `em` units, but they offer fine-grained control

Apologies for the difficult-to-read diff in the meeting pages.

### DIFF
--- a/assets/css/all.sass
+++ b/assets/css/all.sass
@@ -57,3 +57,8 @@ a
 .question
   font-size: 1.2em
   font-weight: bold
+
+.month
+  margin-top: 0.4em
+  margin-bottom: 0.4em
+  color: #606060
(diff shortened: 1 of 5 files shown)


jnewbery:
Thanks Jon. I've taken all of these commits in #79. I think this PR can be closed.

PR #88, opened 2019-12-09
layout: include time and place on PR page

pinheadmz:
I sent this week's meeting link to some coworkers, they asked me how to join the meeting!
This PR adds the where & when to each meeting page:

#### Current

Screen Shot 2019-12-09 at 10 13 44 AM

#### Updated

Screen Shot 2019-12-09 at 10 13 38 AM

PR #89, opened 2019-12-09
Site settings for meeting time and location

jnewbery:
The time and location appear in two places in the home page. Make them site settings so if we need to change them in future, we only have to change one place.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -8,7 +8,10 @@ sass:
   sass_dir: assets/css
   style: :compressed
 
-# Customise atom feed settings (this is where Jekyll-Feed gets configuration information)
-title:		           "Bitcoin Core PR Review Club"
-description:	       "A weekly review club for Bitcoin Core PRs"
-author:              "John Newbery"
+# Site settings
+title:             "Bitcoin Core PR Review Club"
+description:       "A weekly review club for Bitcoin Core PRs"
+author:            "Bitcoin Core PR Review Club contributors"
+meeting_time:      "18:00 UTC"
+meeting_day:       "Wednesday"
+meeting_location:  "#bitcoin-core-pr-reviews IRC channel on <a href=\"https://webchat.freenode.net\">freenode</a>"
--- a/index.md
+++ b/index.md
@@ -5,7 +5,7 @@ title: home
 ### A weekly review club for Bitcoin Core PRs
 
 <span class="question">What is this?</span> &nbsp;A weekly club for reviewing
-Bitcoin Core PRs at **18:00 UTC on Wednesdays** on IRC.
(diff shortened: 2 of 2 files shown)


index.md
```diff
@@ -5,7 +5,7 @@ title: home
 ### A weekly review club for Bitcoin Core PRs
 
 <span class="question">What is this?</span> &nbsp;A weekly club for reviewing
-Bitcoin Core PRs at **18:00 UTC on Wednesdays** on IRC.
+Bitcoin Core PRs at **{{ site.meeting_time }}** on IRC.
```
jonatack: Hacky but perhaps `Bitcoin Core PRs at **{{ site.meeting_time }}s** on IRC.` to keep "Wednesdays".

jonatack: (or define a `general_meeting_time` and a `meeting_time` in the site settings)

_config.yml
```diff
@@ -8,7 +8,9 @@ sass:
...
-# Customise atom feed settings (this is where Jekyll-Feed gets configuration information)
-title:		           "Bitcoin Core PR Review Club"
-description:	       "A weekly review club for Bitcoin Core PRs"
-author:              "John Newbery"
+# Site settings
+title:           "Bitcoin Core PR Review Club"
+description:     "A weekly review club for Bitcoin Core PRs"
+author:          "Bitcoin Core PR Review Club contributors"
+meeting_time:    "18:00 UTC on Wednesday"
+irc_location:    "#bitcoin-core-pr-reviews IRC channel on [freenode](https://webchat.freenode.net/)"
```
pinheadmz: I could be wrong about this but I think markdown is only parsed/rendered in `{{content}}` sections? So for example the link will not look correct if I stick this string in the top of the page in https://github.com/bitcoin-core-review-club/bitcoin-core-review-club.github.io/pull/88

jonatack: `{{content}}` works but in an .html page, markdown does not... run `make preview` and open `http://localhost:4000/16702.html` to test... but you can use CSS/HTML markup if needed. I don't think you need any here though.

jnewbery: Ah yes. Thanks. I've replaced this with html and it seems to work.

index.md
jnewbery: I'm ok with losing the s :)

jnewbery:
Changed the site settings to `meeting_time`, `meeting_day` and `meeting_location`. That means they should be more easily usable in #88

index.md
jnewbery: I've added the s back in!

pinheadmz:
ACK @ 76fd82c7cbe49b0f0600b2f0dd43b3dd109cd878 update to #88 incoming.

jnewbery:
Merged in #88

PR #94, opened 2019-12-12
Add n17639 and update rake posts:new task

jonatack:
to handle PR titles containing double quotes, like for next week's PR 17639, without making Jekyll halt. I went with simply stripping the double quotes. Alternatively, we could convert the double quotes to single quotes. I didn't find it as aesthetic in this case, but it could be an option.

Also made a minor fixup in yesterday's meeting notes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -147,7 +147,7 @@ def create_post_file!(filename, response, date, host)
     line.puts '---'
     line.puts 'layout: pr'
     line.puts "date: #{date}"
-    line.puts "title: \"#{response['title']}\""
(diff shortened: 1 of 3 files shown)


jonatack:
Force-pushed to make this change:
```diff
-title: "tests: Add make check-valgrind to run the unit tests under Valgrind"
+title: "Add make check-valgrind to run the unit tests under Valgrind"
```

Todo: handle auto-removing whitelisted components from the title.

jnewbery:
Looks good. Thanks @jonatack!

PR #97, opened 2019-12-18
Add meeting log for n17639

jonatack:
Thought: Feel free to close this PR if you prefer to push the log from your local timezone to maintain meeting time consistency.


jnewbery:
I don't think we need to have time zones consistent (and if we did we should standardise on UTC)

jonatack:
Ok, and we could script that change if desired. Thanks for tweeting about the meeting log. The holiday break message and see you on Jan 8th was a nice touch.

PR #115, opened 2020-01-24
use docker to isolate host from ruby horror

jamesob:
Installing ruby toolchains always gives me indigestion, so I've added a few
recipes to the Makefile that allow users to work on this site without having to
actually configure Ruby locally.


jonatack:
Concept ACK. Mind adding a docker section to the README?

PR #128, opened 2020-02-10
Remove outdated or bad links

jonatack:
or ones I'd rather not leave up.


jnewbery:
I don't think we should go back and edit previous meeting logs. If that means that old links break, so be it.

PR #140, opened 2020-03-04
Add _posts/2020-03-18-#18234.md

MarcoFalke:


MarcoFalke:
why no netlify?

jnewbery:
Thanks @MarcoFalke , and thanks @jonatack for the additional notes. Jon - perhaps you could add a commit to this to add your suggested references.

I'm going to hold off on merging this for now. This sounds like a great PR to cover, but I may already have a host for March 18th, so this would need to be scheduled to a different date.

jonatack:
That PR was merged quickly! Might be still good to use these notes somehow.

MarcoFalke:
PR has been merged, we should focus on more interesting outstanding stuff

jnewbery:
It's ok to review a PR that's already been merged (the goal of review club is to be educational for participants, not to help PRs get merged), but if you're no longer interested in hosting this, then it makes sense to close.

Thanks for writing notes, perhaps we can get you on the schedule for March 25th?

PR #147, opened 2020-03-17
Add auto logs markup plugin

### LINKED ISSUE #143: add id attributes to meeting logs
I've summarized a previous Review Club meeting for the Optech newsletter (here: https://github.com/bitcoinops/bitcoinops.github.io/pull/366#issuecomment-596774687). For those summaries, it'd be really useful to be able to link specific lines in the meeting log. For that we'd need id or name attributes for each line.

@ajttowns bot does that automatically, eg http://www.erisian.com.au/bitcoin-core-dev/log-2020-03-10.html#l-3. Perhaps it's time to start doing something a bit smarter than copy-pasting meeting logs so we can add more structure to the logs.

jonatack:
This PR creates a plug-in to add line number links (`#l-number`) and style tag markup for the lineno, time, nick, and message fields of the logs.

Further improvments by @jnewbery as described in https://github.com/bitcoin-core-review-club/bitcoin-core-review-club.github.io/pull/148#issuecomment-600392394:
- line number is clickable to get a link to that line
- background highlighting works as you'd expect when the line is targeted
- line numbers are unselectable, so users can copy-paste sections of the log without including the line numbers
- log lines wrap after the timestamp
- logs are html-escaped (take a look at line 255 here: https://deploy-preview-147--bitcoincore-reviews.netlify.com/17428.html to see why that's required)

This PR also removes the code block markdown in the posts in order for the HTML to function.

Closes #143.

Co-authored-by: John Newbery <john@johnnewbery.com>

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -175,8 +175,6 @@ def create_post_file!(filename, response, date, host)
     line.puts "\n"
     line.puts "<!-- TODO: uncomment and add meeting log"
     line.puts "## Meeting Log\n"
-    line.puts "```\n"
-    line.puts "```\n"
     line.puts "--->\n"
   end
 end
--- a/_plugins/auto_logs_markup.rb
+++ b/_plugins/auto_logs_markup.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+# Regex to select lines starting with "HH:MM " time.
+HH_MM = /^([0-1][0-9]|[2][0-3]):[0-5][0-9] .*/
+
+# Regex to select IRC <nick>.
+IRC_NICK = /<.+?>/
+
+# Regex to select "<" and ">" chars.
+LT_GT = /[<>]/
+
+# Maximum digits to display for log lines.
+LINE_DIGITS = 3
+
+# Length of timestamp string being used.
+TIME_SIZE = 'HH:MM'.size
+TIME_SIZE_PLUS_1 = TIME_SIZE + 1 # Micro-perf optimization
+
+NON_BREAKING_SPACE = '&nbsp;'
+
+# Convert logs from plain text to HTML with line number links.
+#
+Jekyll::Hooks.register :documents, :pre_render do |post|
+
+  # Loop through each line of the meeting logs.
+  post.content.gsub!(HH_MM).with_index(1) do |line, index|
+
+    # Separate the log line into useful parts.
+    lineno  = "#{NON_BREAKING_SPACE * (LINE_DIGITS - index.to_s.size)}#{index}"
+    time    = line[0..TIME_SIZE]
+    name    = IRC_NICK.match(line).to_s
+    nick    = name.gsub(LT_GT, '').strip
+    message = CGI.escapeHTML(line[TIME_SIZE_PLUS_1 + name.size..-1])
+
+    # Return the log line as HTML markup.
+    "<table class='log-line' id='l-#{index}'>" \
+      "<tr class='log-row'>" \
+        "<td class='log-lineno'><a href='#l-#{index}'>#{lineno}</a></td>" \
+        "<td class='log-time'>#{time}</td>" \
+        "<td>" \
+          "<span class='log-nick'>&lt;#{nick}&gt;</span>" \
+          "<span class='log-msg'>#{message}</span>" \
+        "</td>" \
+      "</tr>" \
+    "</table>"
+  end
+end
--- a/_posts/2019-05-01-#15557.md
+++ b/_posts/2019-05-01-#15557.md
@@ -9,7 +9,6 @@ status: past
 
 ## Meeting Log
 
-```
 13:01 <@jnewbery> Hi folks!
 13:01 < harding> Hi!
 13:01 < dmkathayat> Hello!
@@ -237,4 +236,3 @@ status: past
 14:01 < RubenSomsen> Thanks John :)
 14:01 <@jnewbery> Feedback is most definitely welcome. Also let me know if you have suggestions for PRs to cover. The aim is for manageable sized PRs (ie 100-150 LOC change seems about right), not too much contextual knowledge required, and trying to cover all the different components
 14:02 <@jnewbery> Feel free to leave feedback in this IRC channel. I'll be monitoring through the week
-```
--- a/_posts/2019-05-08-#10823.md
+++ b/_posts/2019-05-08-#10823.md
@@ -9,7 +9,6 @@ status: past
 
 ## Meeting Log
 
-```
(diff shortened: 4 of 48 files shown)


jnewbery:
This is great. Thanks @jonatack! I'll play around with styles tomorrow.

jnewbery:
Oh dear. This increases build time on my machine from 1-2 seconds to ~12-15 seconds. Similar on netlify.

I'd really prefer to avoid that if possible. 15 seconds isn't the end of the world, but it really slows down making and testing small changes to the site.

jonatack:
FWIW the first plug-in that removes all the ``` markdown should be replaced with a script, because sometimes we use and need that markdown. It's just here for testing the PR.

jnewbery:
> FWIW the first plug-in that removes all the ``` markdown should be replaced with a script

I agree. We could very easily do that in one commit.

jonatack:
Yes, the regex is slow and can probably be optimised. Alternatively, a rake task that converts the post file.

jnewbery:
The slow part seems to be the code that removes the code tag. Deleting that function brings build times back down to 1-2 seconds

jonatack:
Also, upgrading to Ruby 2.7 should yield some speedup... but over time with more logs to handle the build time isn't going to get faster.

jonatack:
> The slow part seems to be the code that removes the code tag. Deleting that function brings build times back down to 1-2 seconds

Which is odd, but you're right. I added that at the end, and things slowed down. It's late and I'll check in tomorrow.

jnewbery:
I implemented some suggested changes in #148. Feel free to add them here.

jonatack:
@jnewbery this should now be RFM or any other changes you see.

harding:
I think you want to do this as a callable filter rather than a globally-applied regex.  E.g. like https://github.com/bitcoin-core/bitcoincore.org/blob/master/_plugins/githubify.rb which is used on files like https://raw.githubusercontent.com/bitcoin-core/bitcoincore.org/master/_releases/0.19.1.md (see `{% githubify https://github.com/bitcoin/bitcoin %}`) to produce content like the PR links on https://raw.githubusercontent.com/bitcoin-core/bitcoincore.org/master/_releases/0.19.1.md

Making this a filter should simplify the code (because there will be fewer false positive matches to deal with) and should make it faster (because it'll be processing much less text overall).

jnewbery:
I've just thought of another enhancement that would be nice: rendering URLs in the meeting logs as links. Doesn't have to be done in this PR. Could be a follow-up.

jnewbery:
@harding - if we do it as a callable filter, is it possible to only run it on sections of a post? We really only want to do it on the part of the post under the "Meeting Log" header (although it's unlikely there'd be any false matches in other parts of the document)

jonatack:
Thanks @harding. At first glance the marginal benefit vs dev time doesn't seem compelling:

- false positives due to lines starting with an `HH:MM` time "should" (:tm: :copyright:) be rare to nonexistent
- IIUC the script needs to be called from somewhere with `{% githubify %}`, e.g. before each meeting log
- the meeting logs represent a majority of the post content to parse anyway: 80%, 83%, and 76% of lines in the last 3 sessions

Maybe I'm missing something. My guess is that if speed is an issue, doing the conversion offline would make the most difference, e.g. a command line rake task called once before pushing a meeting log update.

jonatack:
> I've just thought of another enhancement that would be nice: rendering URLs in the meeting logs as links. Doesn't have to be done in this PR. Could be a follow-up.

Great idea. It could probably be added into the same plugin with a handful of additional lines in a follow-up, once the approach for this one is decided.

jonatack:
Rebased after today's meeting log addition. I might be able to remove one of the regex calls, if yes, will push a commit in a few minutes.

harding:
In my suggestion, you would wrap the meeting log in the filter.  E.g.

```md
# Title

Intro

## Questions

Foo

## Log

{% irc %}
01:23 <harding> bar
{% endirc %}

```

> At first glance the marginal benefit vs dev time doesn't seem compelling

You're probably right, but using a global hook to modify specific sub-content doesn't feel right to me.  That's not an objection, just a note.

jonatack:
Got rid of the second `sub(/<.+?>/, '')` call (when parsing the `message` var). In theory it should be slower than the string#length and integer addition operation that replace it. Probably no discernable difference :)

@harding thanks again and for confirming how the filter is used. I do empathize with the idea.

jnewbery:
I've implemented @harding's suggestion to do this as a tag here: https://github.com/bitcoin-core-review-club/bitcoin-core-review-club.github.io/tree/pr147.2, but it actually seems to be quite a bit slower than doing this as a hook (~6s to render the site compared to <2s for this approach)

harding:
@jnewbery 

> it actually seems to be quite a bit slower than doing this as a hook (~6s to render the site compared to <2s for this approach)

That certainly surprises me!  I took a quick look at the code and don't see anything amiss.

jonatack:
Thanks @jnewbery for implementing this. On my machine the callable filter is a bit slower (5 versus 6 seconds).

Would have to flame graph or do other analysis to know what is optimizable.

That said, in Ruby, call instantiations, object allocations, calling `super`, and garbage collection are often the expensive areas. On large apps, I sometimes adjusted GC settings related to new vs old object lifetimes based on app profile for performance. Not often, but it can be done.

It would be possible to tweak the code inside the loop for a marginal gain at the cost of elegance/syntactic sugar (e.g. not using `with_index` would save a few object allocations), but likely not for a perceptible gain.
 
Upgrading from Ruby 2.5 to 2.7 may also help; 2.5 is more than 2 years old and much of the focus since has been on perf improvements.

jonatack:
Added a micro perf tweak and hoisted the remaining values to constants with code comments.

jnewbery:
I'd left in a commented-out line in the .sass file. I've force pushed a branch that removes that.

Looks RFM to me. Thanks for working on this @jonatack and thanks for review and input @harding !

jnewbery:
gah. I didn't pull the most recent branch before force-pushing my change. I don't think there's a way for me to recover the commit 2fba7558b69cb1524f620015aeadfd9d7b4f666c to undo that change. @jonatack - can you force-push it back? Sorry!

jonatack:
Re-pushed!

jnewbery:
Thanks Jon. I've managed to force-pushed my change without clobbering your latest branch.

jnewbery:
> That certainly surprises me! I took a quick look at the code and don't see anything amiss.

@harding - I was wrong. Not sure what I was building with yesterday, but now that this is merged into master it takes ~6s to build. I think both implementations have similar performance.

PR #150, opened 2020-03-18
Add chart of coin combinations

amitiuttarwar:
Added an appendix section with a chart of all the possible combinations between spent, FRESH & DIRTY and some small notes on validity of combination & how the state could occur.


jnewbery:
Thanks! I've made some small changes and moved the chart to above the meeting log so it's more visible.

One thing I'd suggest to make the table clearer is to add column headings (validity, spentness, freshness, dirtyness, explanation).

amitiuttarwar:
thanks! looks good pulled above :) 

I didn't add column headings because they seem pretty self-apparent (and the first column would have to be wider and less aesthetically pleasing 😛), but I can add if you'd like.

jnewbery:
> I didn't add column headings because they seem pretty self-apparent (and the first column would have to be wider and less aesthetically pleasing 😛), but I can add if you'd like.

ok. It was just a suggestion.

I've made another small tweak and I'll merge now.

Thanks for this contribution!

jonatack:
Great idea, @amitiuttarwar! :100:

PR #151, opened 2020-03-19
Add meetings by PR number

jonatack:
Initial proposal:
- displayed in descending order
- separated by groups of PR number / 1000
- links menu placement and styling TBD

### DIFF
--- a/_includes/linkers/meetings-pages.md
+++ b/_includes/linkers/meetings-pages.md
@@ -8,6 +8,12 @@
   {% assign _index_links = _index_links | append: "<a href='/meetings/'>by date</a>" %}
 {% endif %}
 
+{% if path == "meetings-prs" %}
+  {% assign _index_links = _index_links | append: " | <strong>by PR number</strong>" %}
+{% else %}
+  {% assign _index_links = _index_links | append: " | <a href='/meetings-prs/'>by PR number</a>" %}
+{% endif %}
+
 {% if path == "meetings-components" %}
   {% assign _index_links = _index_links | append: " | <strong>by component</strong>" %}
 {% else %}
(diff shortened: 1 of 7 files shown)


jnewbery:
Thanks for this, Jon!

It seems a little inconsistent to me that for all the other indexes, the whole PR number + title is the anchor, whereas for this the PR number is the anchor and the title is plaintext. Have you tried making the PR number + title the anchor?

I also think we should remove the "<date> meeting" at the end of the line. I don't think it adds much and makes the page more cluttered.

jonatack:
> It seems a little inconsistent to me that for all the other indexes, the whole PR number + title is the anchor, whereas for this the PR number is the anchor and the title is plaintext. Have you tried making the PR number + title the anchor?

Yes, it isn't consistent. I started with styling it like the other pages, and at the time didn't feel it was working well for this case (with no leading date), and ended up with this to see what you think. Will look at it with fresh eyes a bit later.

> I also think we should remove the " meeting" at the end of the line. I don't think it adds much and makes the page more cluttered.

I agree that it's clutter. My thought was that otherwise people might think it's the PR filing date and not the meeting date, but happy to remove it.

jnewbery:
> I also think we should remove the " meeting" at the end of the line. I don't think it adds much and makes the page more cluttered.

Oops. I meant 'remove the `<date> meeting` from the end of the line. Github removed the thing that looked like a tag. I think there should be no date at all.

jnewbery:
I've pushed a commit with some suggested changes to make the style more consistent with the other indexes.

jnewbery:
... and rebased on master

jonatack:
ACK

Subjective of course, but as a general follow-up, (I think) I'd prefer the following for all the meeting listings: 
- Only the PR number in the link tag, and the PR name in white text; this seems more calm and readable and focuses the choice of link on the PR number
- Perhaps without the `#` and `>>` as well, which seems cleaner to me

jnewbery:
> the PR name in white text

This doesn't work for the 99% of people who have a white background :)

jnewbery:
Thanks for the review Jon. I've squashed the commits more logically.

jnewbery:
> as a general follow-up, (I think) I'd prefer...

Let's do this in a separate PR. I think the most important thing is for the index styles to be consistent. Now that this is merged, we can iterate on the style in a follow-up PR.

PR #153, opened 2020-03-19
Convert URIs in meeting logs to HTML links

jonatack:
Adds 1 or 2 seconds to the build but looks great. Whitelisted the uri schemes, otherwise any word ending with a colon was being turned into a link. Nevertheless, much cleaner to use the Ruby URI standard library than futzing around with regexes for this.


jnewbery:
Build time for me is up to 8 seconds, which is tolerable. Logs look fantastic. Thanks @jonatack !

PR #154, opened 2020-03-19
Don't clean jekyll between previews

jnewbery:
No need to clean jekyll when doing a `make preview`. Just run jekyll serve --incremental so we don't rebuild everything.

This brings build times down from ~8s to <0.5s.

`make build` still does a clean since it's not doing an incremental build so we might as well clean. Also add a `make clean` target.

Historic note: I copied our makefile from optech, and Dave copied that makefile from bitcoin.org, where he needed to use clean because some of their custom plugins didn't work well with incremental builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 all: build
 
-preview:
+clean:
 	bundle exec jekyll clean
+
+preview:
 	bundle exec jekyll serve --future --drafts --unpublished --incremental
 
 build:


jonatack:
Great! This clears up the mystery :smile:

IIUC correctly, for testing the plugin, would call `make clean && make preview` and otherwise `make preview` would be enough.

harding:
@jonatack 

> IIUC correctly, for testing the plugin, would call make clean && make preview

FYI, you can call make with multiple targets and it'll run them in series (stopping at the first failure, like normal).  E.g. `make clean preview` ; it doesn't matter here, but this is handy for more complicated makefiles as make will only run any common dependencies of the two targets once in the single-command version whereas it would run the common dependencies twice if you use the two commands in succession.

jonatack:
> FYI, you can call make with multiple targets and it'll run them in series (stopping at the first failure, like normal). E.g. `make clean preview` ; it doesn't matter here, but this is handy for more complicated makefiles as make will only run any common dependencies of the two targets once in the single-command version whereas it would run the common dependencies twice if you use the two commands in succession.

TIL, thanks @harding!

PR #155, opened 2020-03-19
Clean up links in meeting logs

jonatack:
Found a few cases where a link could be broken if followed by trailing punctuation or brackets, for example:
- https://bitcoincore.reviews/16981.html#l-135
- https://bitcoincore.reviews/16939.html#l-46

and also see a `file:` turned into a useless link, so removed "file" from the schemes, as well as 'ftp' and 'mailto' until we see those actually needed.

### DIFF
--- a/_plugins/auto_logs_markup.rb
+++ b/_plugins/auto_logs_markup.rb
@@ -3,7 +3,10 @@
 require 'uri'
 
 # URI schemes to accept for extraction.
-URI_SCHEMES = %w(file ftp http https mailto).freeze
+URI_SCHEMES = %w(http https).freeze
+
+# Trailing chars to remove from URIs.
+TRAILING = /[[:punct:]]+$/
 
 # Regex to select lines starting with "HH:MM " time.
 HH_MM = /^([0-1][0-9]|[2][0-3]):[0-5][0-9] .*/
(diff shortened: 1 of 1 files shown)


jnewbery:
Good catch! Thanks Jon

jonatack:
I think this is ready.

jonatack:
Review for review now.

jonatack:
A final tweak to improve appearance of the links :)
```diff
-      message.sub!(uri, "<a href='#{uri.gsub(TRAILING, '')}' target='blank'>#{uri}</a>")
+      message.sub!(uri, "<a href='#{uri.gsub!(TRAILING, '')}' target='blank'>#{uri}</a>")
```

jonatack:
I think this is good to merge since it fixes issues and is now a minimal change.

jnewbery:
There are a couple of problems with this:

- it removes the trailing punctuation, rather than just excluding it from being part of the link (which I think is what we want)
- it removes the trailing `/` character, which in some cases might break the link (the trailing slash might be part of the URL - see for example the difference between https://jonatack.github.io/articles/on-reviewing-and-helping-those-who-do-it/ and https://jonatack.github.io/articles/on-reviewing-and-helping-those-who-do-it)

jonatack:
> * it removes the trailing punctuation, rather than just excluding it from being part of the link (which I think is what we want)

Fixed with the follow-up commit (I think).

> * it removes the trailing `/` character, which in some cases might break the link (the trailing slash might be part of the URL - see for example the difference between https://jonatack.github.io/articles/on-reviewing-and-helping-those-who-do-it/ and https://jonatack.github.io/articles/on-reviewing-and-helping-those-who-do-it)

True, will modify the regex to exclude touching that character.

PR #157, opened 2020-03-20
Add variable nick colors to logs

### LINKED ISSUE #156: Color-code meeting participants
We render all meeting participants in the same color. A couple of people have suggested to me that each participant should have a separate color to make it easier to follow threads. Not sure how much effort it'd be to implement it or how useful in general, but I'm opening this issue to track that request.

currently:

image

proposed:

image

jonatack:
Closes #156. Color choice TBD :)

### DIFF
--- a/_plugins/auto_logs_markup.rb
+++ b/_plugins/auto_logs_markup.rb
@@ -31,12 +31,23 @@ module Jekyll
 
     NON_BREAKING_SPACE = '&nbsp;'
 
+    COLORS = %w(brown goldenrod cadetblue chocolate cornflowerblue coral crimson
+      forestgreen darkblue firebrick blue green grey hotpink indianred indigo
+      blueviolet maroon mediumblue mediumpurple mediumseagreen navy fuchsia
+      olive orchid purple red seagreen sienna orange slateblue peru salmon teal
+      magenta steelblue rebeccapurple tomato violet darkcyan).freeze
+
+    NUM_COLORS = COLORS.size.freeze
+
     def initialize(tag_name, text, tokens)
       super
     end
 
     def render(context)
       output = super
+
+      # Reset color data for each post. Seed the index with part of the content.
+      colors, color_index = {}, (output[-256..-250] || '').bytes.sum
 
       # Loop through each line of the meeting logs.
       output.gsub!(HH_MM).with_index(1) do |line, index|
@@ -46,6 +57,8 @@ module Jekyll
         time    = line[0..TIME_SIZE]
         name    = IRC_NICK.match(line).to_s
         nick    = name.gsub(LT_GT, '').strip
+        color   = colors[nick] || (color_index = (color_index + 1) % NUM_COLORS ;
+                                   colors[nick] = COLORS[color_index])
         message = CGI.escapeHTML(line[TIME_SIZE_PLUS_1 + name.size..-1])
 
         # Extract URIs from the message and convert them to HTML links.
@@ -60,7 +73,7 @@ module Jekyll
             "<td class='log-lineno'><a href='#l-#{index}'>#{lineno}</a></td>" \
             "<td class='log-time'>#{time}</td>" \
             "<td>" \
-              "<span class='log-nick'>&lt;#{nick}&gt;</span>" \
+              "<span class='log-nick' style='color:#{color}'>&lt;#{nick}&gt;</span>" \
               "<span class='log-msg'>#{message}</span>" \
             "</td>" \
           "</tr>" \


_plugins/auto_logs_markup.rb
```diff
@@ -38,6 +48,7 @@
     time    = line[0..TIME_SIZE]
     name    = IRC_NICK.match(line).to_s
     nick    = name.gsub(LT_GT, '').strip
+    color   = name_colors[nick] || (name_colors[nick] = colors[color_index += 1])
```
jonatack: Note that the colors array is currently 63 elements in size; if we have more than that many participants in a post, the color will become `nil`. We could check that `color_index < colors.size` and if true reset the index to 0, or ensure the color array is large enough.

jonatack: Note that the colors array is currently 63 elements in size; if we have more than that many participants in a post, the color will become `nil` and fall back to the color (green) set in the `.log-nick` CSS class in all.sass. I don't think it's an issue, but we could check that `color_index < colors.size` and if true reset the index to 0, and/or ensure the color array is large enough (which also avoids anyone having the same color). For now I kept this to a one-liner in favor of a large color array.

_plugins/auto_logs_markup.rb
```diff
@@ -52,7 +63,7 @@
         "<td class='log-lineno'><a href='#l-#{index}'>#{lineno}</a></td>" \
         "<td class='log-time'>#{time}</td>" \
         "<td>" \
-          "<span class='log-nick'>&lt;#{nick}&gt;</span>" \
+          "<span class='log-nick' color='#{color}'>&lt;#{nick}&gt;</span>" \
```
jnewbery:
```suggestion
          "<span class='log-nick' style=\'color:#{color}'>&lt;#{nick}&gt;</span>" \
```

I think this needs to be a style tag. Using an html color attribute doesn't seem to work for me.

_plugins/auto_logs_markup.rb
```diff
@@ -26,9 +26,19 @@
...
+magenta maroon mediumblue mediumpurple mediumseagreen navy olive orange orchid
+papayawhip peachpuff peru pink plum purple rebeccapurple red rosybrown royalblue
+salmon seagreen seashell sienna silver skyblue slateblue springgreen steelblue
+tan teal thistle tomato turquoise violet wheat yellow yellowgreen).freeze
+
 # Convert logs from plain text to HTML with line number links.
 #
 Jekyll::Hooks.register :documents, :pre_render do |post|
+  # Reset color data for each post.
+  colors, color_index, name_colors = COLORS.shuffle, -1, {}
```
jnewbery: Why shuffle? I'd prefer to keep this deterministic.

jonatack: The idea was so people aren't stuck with the same color in every post.

_plugins/auto_logs_markup.rb
jonatack: Thanks -- done.

_plugins/auto_logs_markup.rb
jonatack: The idea was so people aren't stuck with the same color in every post, and also, the colors would be based on order of posting.

jonatack:
>     * 63 is plenty of colors. I actually think 16 would be enough. We might have more than that many people say hi at the start, but there are unlikely to be >16 active participants.

For now ;) I like the wide variety but did try to avoid the clearer colors both for contrast on white backgrounds and not blinding people using dark mode.
 
>     * How about we just apply a class attribute `log-nick-color1`, `log-nick-color2`, etc, in the plugin and then define CSS rules for those classes in the CSS file? That seems like a more natural separation to me.

I see your point, but didn't want to mess with the CSS file and that seems like more hard-coding. Maybe someone can do it as a follow-up.

_plugins/auto_logs_markup.rb
jonatack: The idea was so people aren't stuck with the same color in every post, and the colors would be based on order of posting in the first post parsed, and also the cumulative number of participants is higher, so if we don't reset, some colors might be used by more than one participant in a meeting.

jonatack:
Remove a couple more colors that seemed too light (peachpuff and seashell).

amitiuttarwar:
THIS IS SO EXCITING!

jnewbery:
I've implemented my suggestions here: https://github.com/jnewbery/bitcoin-core-review-club.github.io/tree/pr157.1. I have quite a strong preference for putting any styles in a CSS class, and also a preference for builds being deterministic. One reason we'd want that is it means we can easily diff a branch's build against the master build:

```
> git co <branch>
> make
> mv _site _site.<branch>
> git co master
> make
> diff _site _site.<branch>
```

With non-deterministic builds, there's a lot of noise in those diffs.

jonatack:
Thanks @jnewbery, great to have the code to look at and test. For me either way works; here would be my feedback on the second approach compared to the first.

Benefits:
- deterministic... so far I have diffed the codebase but not the website, but the value is clear
- the color data is in the CSS file, but at a cost (last point below)

Drawbacks:
- fewer colors (maybe 24? easy to change)
- some sequentially neighbouring colors are quite similar to each other (easy to change)
- the same colors are seen first... could a deterministic offset be added? Say, PR number % number of colors  
- the source HTML no longer provides direct feedback on which color is used
- the color logic is shared between the code and the CSS, e.g. the color class naming and the modulo value, which depends on the number of color classes in the CSS file (and which I'd suggest hoisting to a constant); so a slightly more brittle and potentially error-prone to change

jnewbery:
Thanks for looking, Jon. For your feedback:

> fewer colors (easy to change)

I don't think more than 16 colors are needed, since I don't think it's possible for more than that many people to be involved in threads simultaneously without it becoming impossible to parse. We can increase this number if we need to though.

> some sequentially neighbouring colors are quite similar to each other (easy to change)

We can fix the colors so this doesn't happen. I think that's an advantage of my approach (if you're choosing the colors randomly you can't guarantee this won't happen).

> the same colors are seen first... could a deterministic offset be added? (easy to change; initialize the index to the PR number instead of -1)

Does it matter?

> the source HTML no longer provides direct feedback on which color is used

This is best practice for website design - the structure goes in the html and the style goes in the CSS. It's better not to mix the two.

> the color logic is shared between the code and the CSS, e.g. the color class naming and the modulo value, which depends on the number of color classes in the CSS file (and which I'd suggest hoisting to a constant); so a slightly more brittle and potentially error-prone to change

Again, this is how websites usually work. The style is added to an element by giving it a class, and that class is defined in the CSS file. We do the same for `log-row`, `log-lineno` and `log-time` for example.

jonatack:
Added a commit to make it determistic seeded by the PR number. That addresses points 1 and 3. Having a snack, then will look at separating similar colors (point 2). As for the CSS, I know it's the standard practice, but in this case I think the benefits are slim to negative compared to having all the logic in one place with a color array that is easy to change and keep large/diverse for the fun factor and count the size of... at least while looking at separating the colors, then if you still want to hardcode it into CSS go for it ;)

jonatack:
Addressed points 1-3. This keeps a larger palette from post to post for the fun and diversity, spreads out the similar colors (I'm no color or front end expert though), while nevertheless making it deterministic.

jnewbery:
Needs rebase on master now that #159 is merged :grimacing: 

I don't think it matters too much whether we do this in CSS or with a direct style attribute. Let's go ahead with your version since you contributed this PR.

jonatack:
I missed something in the rebase... don't merge yet :)

jonatack:
Rebased, then realised the `post` object is no longer available... looking.

jonatack:
I didn't see a trivial way to get the PR number without passing it in with the tags so changed the index seed to the sum of the ascii values of a few characters of the log.

jnewbery:
image

I've taken a look at one log from the built site and two of the colors (yellow and aqua) are very difficult to read. I haven't looked at more logs.

I really think this would be easier if there were a small palette of colors that we know are legible.

jonatack:
I'll do another review on the colors, this time with white background.

jonatack:
I agree, on a white background some were too light. Removed ~10, here are all the remaining ones, any others you would remove?

Screenshot from 2020-03-23 17-29-12

jnewbery:
Those all look fine to me

jonatack:
Done, squashed a couple of commits.

jonatack:
Feel free to take it from here for anything else you see :)

jnewbery:
Squashed and merged. Thanks @jonatack ! :rainbow:

PR #166, opened 2020-03-24
Add n18401

jnewbery:

PR #167, opened 2020-03-25
n18113: fix link that changed from 200 to 404

jonatack:
While running the tests to check #166, they found a link that just went from 200 to 404 due to a force push in the PR. This kind of fix could arguably be automated at some point.


jnewbery:
Thanks Jon!

Perhaps we should push the branch to bitcoin-core-review-club/bitcoin _before_ the meeting and encourage hosts to link to commits there in order to prevent them from breaking when the PR branch changes?

PR #169, opened 2020-03-31
Fix 18401 future commit links

fjahr:
The commits referenced were outdated. I am pretty sure I got the right ones, but please check again, one of the commits referenced seems to have been split.


jnewbery:
Thanks! I've checked that these links are now correct.

PR #176, opened 2020-04-10
Exclude non-site files from build

jnewbery:
There are some files that get included in the site build that shouldn't be, eg go to https://bitcoincore.reviews/Makefile. Exclude them.


jnewbery:
To test this, run `make clean preview` and verify that none of the excluded files are in the `_site` directory.

jonatack:
Tested ACK.

PR #185, opened 2020-05-07
n16442: add 2020-05-07 meeting log (BIP157)

jonatack:
I don't know if you want to add this, here or somewher else, but I think it's a helpful resource.

(I reset my IRC connection at 19:01 so may have missed a line or two during that minute.)

### DIFF
--- a/_posts/2020-05-07-#18877.md
+++ b/_posts/2020-05-07-#18877.md
@@ -1,0 +1,162 @@
+---
+layout: pr
+date: 2020-05-07
+title: "Serve cfcheckpt requests"
+pr: 18877
+authors: [jimpo, jnewbery]
+components: ["p2p"]
+host: jnewbery
+status: past
+commit: 967e2b10
+---
+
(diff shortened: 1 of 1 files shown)


amitiuttarwar:
imo it makes more sense to add logs as an additional meeting (for 18877), with a header explaining the ad hoc context. easier to find the notes & leaves possibility for any future sessions like this

jonatack:
> imo it makes more sense to add logs as an additional meeting (for 18877), with a header explaining the ad hoc context. easier to find the notes & leaves possibility for any future sessions like this

Thanks @amitiuttarwar. I think I prefer your suggestion too, though either way is fine with me.

jnewbery:
Thanks for this, Jon. I've moved it to its own page as you and Amiti suggested. I agree it looks better that way. Can you take a look and merge if you're happy?

jonatack:
ACK, thanks John and Amiti.

PR #200, opened 2020-05-23
Update GitHub labels used by Bitcoin Core

jonatack:
as per https://github.com/bitcoin/bitcoin/labels

Some of these, like `Data corruption`, `Privacy`, and `Refactoring`, are arguably not codebase components but may be practical or the only label.


jnewbery:
Looks good. Thanks Jon!

PR #202, opened 2020-05-23
CONTRIBUTING.md: use 3rd-person narrative in the notes

jonatack:
...as I did the opposite today when editing fjahr's notes ;)


jnewbery:
Thanks Jon. I think this is a good suggestion.

PR #222, opened 2020-06-27
n18468  component tag fix and notes tidy-ups

jonatack:
Noticed the component tag was creating a second `refactor` component type in https://bitcoincore.reviews/meetings-components. A quick look at the GitHub PR shows that `refactoring` is still the correct tag.

While fixing it up, did a quick grammar and presentation pass through the notes and questions. Don't hesitate to drop that commit.


jnewbery:
Thanks Jon!

PR #223, opened 2020-06-27
Rakefile: generate a (slightly) clearer new post template

jonatack:
When reviewing the latest post, I remember there was a dangling html comment fragment left over from the notes that was causing the log html to be uncommented and visible.

So I generated a new post to have a fresh look and reckon with the current template it wouldn't be difficult to confound the notes html and the log html.

This PR separates and consolidates the two sections a bit more and also describes where to add the meeting log.


jnewbery:
LGTM. Thanks Jon!

PR #252, opened 2020-09-10
n18267: add meeting log

jnewbery:
Unfortunately our markdown parser gobbles multiple consecutive spaces in the meeting log.

I didn't have time to figure out if we can fix that so I included the ascii art as an image.


jonatack:
The image looks great! Thanks, John. I copied over the log last night but was unsure how much of it to include--I think it broke all previous records in length!

PR #257, opened 2020-09-16
Update Gemfiles and README for Ruby 2.6

jonatack:
and add an inoffensive `.ruby-version` file used by some Ruby gem managers like rbenv (https://github.com/rbenv).


jnewbery:
Works for me. Thanks @jonatack 

I had a bit of trouble running `gem update --system && gem update && bundle update` on my rvm-installed ruby 2.6.6 and had to hand-edit the mkdir directory as described here: https://stackoverflow.com/a/63095629/933705. After that, I didn't have any problems.

jonatack:
Thanks!

PR #260, opened 2020-09-30
Update gzhao's username

glozow:
I'm terribly sorry for the inconvenience, but I changed my handles everywhere (this should be the only time I do this), and would really appreciate if the links could be updated. Thank you :)


jnewbery:
LGTM. Thanks @glozow !

PR #264, opened 2020-10-08
Announce meeting on 19953

jnewbery:
Notes and questions to follow

Also adds a new 'taproot' component, so that a list of all the taproot meetings is generated automatically. It's not strictly a component, but that field works well for general tags.


jonatack:
Concept ACK!

jonatack:
Tested ACK

jonatack (commented):
Tested ACK

jnewbery:
Thanks Jon!

PR #272, opened 2020-10-23
Add notes and questions for 20158

dongcarl:


jnewbery:
Thanks Carl. I made some minor style fixups, but otherwise looks great.

PR #281, opened 2020-11-18
Fix two small typos in n20079

satsie:
Fix two small typos in n20079:

- `send` -> `sent`
- `disconecting` -> `disconnecting`


jnewbery:
Thanks @satsie!

PR #284, opened 2020-11-22
Update 2020-11-25-#19858.md

amitiuttarwar:
looks like we talk about eclipse attacks a fair amount =P


jonatack (commented):
ACK

Thanks @amitiuttarwar!

PR #305, opened 2021-02-03
Change earlier edit to addendum to preserve the original, which was r…

chrisguida:
…eferenced in discussion


jnewbery:
Thanks @chrisguida!

PR #310, opened 2021-02-17
Redecorate with dark theme

pinheadmz:
Offering this for late night review club crammers to make the site easier on the eyes. For IRC logs, I switched out a few colors for better contrast against the dark background.

Here's some examples screen shots showing main page, IRC log, and a `code` block...

##

Screen Shot 2021-02-17 at 9 59 48 AM

##

Screen Shot 2021-02-17 at 10 00 00 AM

##

Screen Shot 2021-02-17 at 10 01 33 AM

### DIFF
--- a/assets/css/all.sass
+++ b/assets/css/all.sass
@@ -24,18 +24,27 @@ $large-screen: 1024px
 
 /* This is an example style sheet that comes default - just delete everything after this comment this to start from a blank slate */
 
+html
+  height: 100%
+
+body
+  height: 100%
+  min-height: 100%
+
 #color-mode
   display: none
 
 #color-mode-switch
   user-select: none
-  float: right
+  right: 0
+  position: absolute
 
 // Light (default)
 #color-mode:not(:checked) ~ .site-container
   --bg-color: white
   --text-color: black
   --link-color: #0000f0
+  --link-visited-color: #551b8c
   --code-bg-color: #f5f5f5
   --log-bg-color: #f0f0f0
   --log-color: #808080
@@ -47,6 +56,7 @@ $large-screen: 1024px
   --bg-color: black
   --text-color: white
   --link-color: #58a6ff
+  --link-visited-color: #824bb7
   --code-bg-color: #565656
   --log-bg-color: #565656
   --log-color: #080808
@@ -59,6 +69,9 @@ a
 a
   &:hover
     color: $secondary-color
+a
+  &:visited
+    color: var(--link-visited-color)
 
 .site-container
   max-width: 2250px
@@ -66,12 +79,16 @@ a
   padding: 1rem 2rem 0 2rem
   background-color: var(--bg-color)
   color: var(--text-color)
+  min-height: 100%
+  overflow: auto
 
 .logo
   text-decoration: none
   color: var(--text-color)
   font:
     size: 48px
+  a
+    color: var(--text-color)
 
 .dt-published
   color: #828282
--- a/assets/css/all.sass
+++ b/assets/css/all.sass
@@ -67,11 +67,12 @@ a
   color: var(--link-color)
 
 a
-  &:hover
-    color: $secondary-color
-a
   &:visited
     color: var(--link-visited-color)
+
+a
+  &:hover
+    color: $secondary-color
 
 .site-container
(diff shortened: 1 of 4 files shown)


pinheadmz:
Converting to draft for now, will investigate the simplest way to make this user settable. @jonatack @jnewbery this may introduce javascript to the website, is that a game changer?

jnewbery:
> Converting to draft for now, will investigate the simplest way to make this user settable. @jonatack @jnewbery this may introduce javascript to the website, is that a game changer?

I'd definitely prefer it if there was a way to achieve this without adding javascript.

jonatack:
> I'd definitely prefer it if there was a way to achieve this without adding javascript.

Same, though a quick web search for doing this with jekyll turns up solutions that need a tiny bit of JS to get around the static aspect (often with media queries and localStorage). Maybe with this: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

jnewbery:
@pinheadmz are you still working on this?

pinheadmz:
Well, I haven't given up on this, but you can close if you are just cleaning up the repository. I discovered a trick using HTML and CSS they can actually switch some color variables in other areas of the style. But haven't got it quite to work how I envisioned it yet.

jnewbery:
No need to close this. I just wanted to check that it was still active.

pinheadmz:
Ok, refactored this with a toggle button (💡 emoji floating in top-right corner) and the actual CSS color switching requires NO javascript. However -- getting the user preference to persist does. Users with JS disabled will have to click the button every time they switch pages or reload. All I'm doing in the JS is saving a boolean in the browser local storage.


Mar-30-2021 17-01-06

jnewbery:
Very cool. This looks good to me. Thanks @pinheadmz!

@jonatack - you're a dark mode connoisseur. Could you check that this satisfies your needs (and merge if so)?

jonatack:
Thanks @pinheadmz! Using the Dark Reader extension slows things down and is a privacy hole, so this is great.

Tested and the only issues I see are

- White background at the bottom of the page (see screenshot below)
- Clicked links don't seem to be changing color

Do you want to have a look at these? Otherwise this is :+1:

jonatack:
(tested with the Firefox 87 on Linux)

Screenshot from 2021-03-31 12-14-49

pinheadmz:
@jonatack thanks for taking a look. I think I figured out the nits. Tested on OSX Chrome / Safari / Firefox and iOS Chrome

assets/css/all.sass
```diff
@@ -59,19 +69,26 @@ a
 a
   &:hover
     color: $secondary-color
+a
+  &:visited
+    color: var(--link-visited-color)
```
jonatack, line 74: By inverting these two, ISTM we get the hover action on the visited links just like on the not-yet visited ones.
```diff
-a
-  &:hover
-    color: $secondary-color
 a
   &:visited
     color: var(--link-visited-color)
 
+a
+  &:hover
+    color: $secondary-color
+
```

jonatack (commented):
This is great--thanks for keeping at it! I tested again on Firefox and compared to the production site without and without the dark reader extension. LGTM, only one remaining suggestion.

assets/css/all.sass
pinheadmz: Great idea, added in 1b7d6d520ac056a947a66135aa3bfd979ecdcdce

tbh, I didn't even realize the `.sass` files were top-down hierarchical 🤦

jonatack:
Tested ACK.

The remaining things I looked at:
- The color of the log line links in default (light) mode has changed from grey to the blue for links. I liked the grey but it seems to be more coherent now
- The blue color of the unvisited links is a bit darker now in default (light mode). I played with it briefly and didn't see why.
- The colors in dark mode are really dark compared to dark reader (but it's just habit, not an issue)
- I'm not sure the meaning of the light bulb icon is clear to people on its own. Maybe a "dark mode/light mode" link to the right of the code of conduct one.

If any of these are a concern, they can be done in a follow-up.

Thanks @pinheadmz!

jnewbery:
Thanks zippy!

PR #311, opened 2021-02-18
ellemouton:pr20277

ellemouton:
Announcement along with notes and questions for PR 20277.


jonatack:
Interesting PR and early notes, thanks! Heading out for some sport but will review soon in a few hours.

ellemouton:
note that the PR that this pr references has just changed and no longer contains the "must request tx to receive tx" logic. I think that is being moved to a separate PR. I will update the PR number here once it is published 👍

PR #313, opened 2021-02-19
minisketch math(s) notes

glozow:
basically explains all the math, maybe helpful 🤷 lmk


glozow:
Looks like this

image

jnewbery:
Thanks Gloria. These are great!

PR #320, opened 2021-03-15
Fix a host's handle

amitiuttarwar:
One of the review clubs that lightlike/mzumsande hosted was mapped to the irc handle instead of github.


jonatack (approved):
ACK, thanks!

jnewbery:
Thank you!

PR #328, opened 2021-04-02
Remove max-width from site-container

dergoegge:
Screenshot_2021-04-02 Multiprocess Add basic spawn and IPC support(1)

There is a white edge on large screens when darkmode is enabled. (See screenshot)
(Not sure if this is a concern but this fix would make lines longer on screens that are wider than 2250 pixels.)

### DIFF
--- a/assets/css/all.sass
+++ b/assets/css/all.sass
@@ -75,7 +75,6 @@ a
     color: $secondary-color
 
 .site-container
-  max-width: 2250px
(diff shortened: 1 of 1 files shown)


jonatack:
Thanks! Having a look.

jonatack:
Tested ACK on a 3840x1600 screen. Thanks @dergoegge!

PR #334, opened 2021-04-17
Differentiate the titles of Python Minisketch meetings 1 and 2

jonatack:
as they currently appear the same apart from the date in the meetings listings.


jnewbery:
LGTM. Thanks Jon!

PR #343, opened 2021-05-24
Add max-width attribute to limit page width

jnewbery:
Text is more difficult to read when the columns are very wide. Limit the maximum website width to 1200px to prevent very long lines.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -9,13 +9,11 @@
   <label id="color-mode-switch" for="color-mode">💡</label>
 
   <div class="site-container">
-
-    {% include header.html %}
-
-    {{ content }}
-
-    {% include footer.html %}
-
+    <div class="page-content">
+      {% include header.html %}
+      {{ content }}
+      {% include footer.html %}
+    <div>
   <div>
 
 </body>
--- a/assets/css/all.sass
+++ b/assets/css/all.sass
@@ -82,6 +82,10 @@ a
   min-height: 100%
   overflow: auto
 
+.page-content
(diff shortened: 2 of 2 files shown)


assets/css/all.sass
```diff
@@ -30,6 +30,8 @@ html
 body
   height: 100%
   min-height: 100%
+  max-width: 1200px
+  margin: 0 auto
```
jonatack, line 34: Concept ACK. Testing it reminded me that this does not work as-is with the dark mode (see #328).

jnewbery: Thanks! Should be fixed now.

_layouts/default.html
```diff
@@ -9,13 +9,15 @@
...
-
+    <div class="page-content">
+  
+      {% include header.html %}
+  
+      {{ content }}
+  
+      {% include footer.html %}
+  
+    <div>
```
jonatack, line 20:
```diff
   <div class="site-container">
     <div class="page-content">
-  
+
       {% include header.html %}
-  
+
       {{ content }}
-  
+
       {% include footer.html %}
-  
+
     <div>
   <div>
```

jonatack (commented):
Nice! Tested ACK modulo extra spaces nit

_layouts/default.html
jonatack: (or just drop the extra line breaks too)

jnewbery: done!

jonatack:
ACK per `git diff a30b631 8cd7710` :)

PR #351, opened 2021-06-18
include Question 1 and CONTRIBUTING.md link in template

glozow:
We have this question in every review club and it always requires going to a previous note to copy the text/link... maybe just include it in the TODO comment to save time?


jonatack:
Good idea.  Could even add the numbers for e.g. 10 questions or any other layout that seems needed every time or easier to delete than to add.

jnewbery:
vnice

PR #356, opened 2021-07-03
n22363: small fixes (s/wallet_util.py/script_util.py/)

theStack:
All the helpers used for the refactoring are residing in script_util.py (only the bug is fixed in wallet_util.py, which then again uses script_util.py).

By the way, really nice idea with the script type fill-out-sheet 🚀


glozow (commented):
ACK, thanks for correction 😅

PR #397, opened 2021-09-23
Notes for 22950

amitiuttarwar:
it might be better to have links or something to the images, because they are kinda a lot 😅 

also, it seems like the image quality isn't the best right now. I'm not sure why but I'll try to improve them tomorrow.


jnewbery:
I've pushed some minor updates to the text, and also updated the images so that:

- they're links to the image resource. Clicking on them will open that image in a full page view.
- they're centered on the page, and have a max-width, so they're viewable on smaller devices/displays.

> it seems like the image quality isn't the best right now. I'm not sure why but I'll try to improve them tomorrow.

We can update the files later.

PR #426, opened 2021-12-02
fix n23614 irc log (broken timestamp)

theStack:
Line 50 in the IRC log for n23614 currently looks odd:

image

Correcting the timestamp should hopefully fix that.


glozow:
Oopsie, thanks @theStack !

PR #427, opened 2021-12-03
n23443 notes

naumenkogs:


_posts/2021-12-08-#23443.md
```diff
@@ -12,13 +12,55 @@ commit:
...
+4. What is the overall handshake and "registration for reconciliation" protocol flow?
+
+5. Why Erlay protocol version feature is useful? Under which conditions it should be bumped?
+
+6. What is the reason for generating reconciliation salt (`m_local_salts`)? How is it generated
+   and why?
+
+7. Why does it make sense in some cases to disconnect a protocol violating peer rather than
+   just ignoring the message?
+
```
glozow, line 63: Suggestion, if you have time: a few implementation-related questions may encourage attendees to dig more thoroughly into the PR code. For example, "what does `TxReconciliationTracker::Impl::m_mutex` guard and when does the lock need to be held?" or "why might we use a [`std::unordered_map`](https://en.cppreference.com/w/cpp/container/unordered_map) instead of a [`std::map`](https://en.cppreference.com/w/cpp/container/map) to track salts and states?"

_posts/2021-12-08-#23443.md
```diff
@@ -12,13 +12,55 @@ commit:
...
+4. What is the overall handshake and "registration for reconciliation" protocol flow?
+
+5. Why Erlay protocol version feature is useful? Under which conditions it should be bumped?
+
+6. What is the reason for generating reconciliation salt (`m_local_salts`)? How is it generated
+   and why?
+
+7. Why does it make sense in some cases to disconnect a protocol violating peer rather than
+   just ignoring the message?
```
glozow, line 62: I recommend prompting the attendees to look into the code. Giving them a specific number of things to search for also helps.
```suggestion
7. This PR adds 4 new conditions under which nodes will disconnect a peer. What are they?
When does it make sense to disconnect a peer rather than just ignore the message?
```

_posts/2021-12-08-#23443.md
```diff
@@ -12,13 +12,55 @@ commit:
...
 
-<!-- TODO: Before meeting, add notes and questions
 ## Notes
 
+* Erlay is a new transaction relay protocol which is supposed to reduce the bandwidth nodes use to
+  announce transactions.
+
+  - We looked at Erlay at the review club [previously](https://bitcoincore.reviews/18261).
+
+  - Now, we will review a particular subset of Erlay commits to be merged first.
```
glozow, line 22: Assuming this line can be removed, since it has its own bullet point

glozow (commented):
Thanks @naumenkogs! A few suggestions

_posts/2021-12-08-#23443.md
naumenkogs: I couldn't think of better questions :)

glozow:
Added some very minor edits. Thanks @naumenkogs, looking forward to the meeting! :)

PR #430, opened 2021-12-08
announce n10102

glozow:


glozow:
cc @ryanofsky, thanks for agreeing to host! :)

No rush on the notes and questions btw, anytime before December 17 would be perfect.

PR #434, opened 2021-12-10
Add 23724

0xB10C:
Might add more questions in a follow-up!

### DIFF
--- a/_posts/2021-12-15-#23724.md
+++ b/_posts/2021-12-15-#23724.md
@@ -13,22 +13,41 @@ commit:
 
 ## Notes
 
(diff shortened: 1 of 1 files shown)


glozow (commented):
Vielen danke @0xB10C, die Notes und Fragen sind super.

As a build systems noob, I might suggest some basic questions (even if the answer is literally in the Notes) such as:

- What happens if the sys/sdt.h header is not present during compilation?
- For guix builds, why do we need to add `systemtap` as a dependency instead of using whatever binary the user has built?
- What triggers the setting of `ENABLE_TRACING`? When is this set?
- What USDT-related things happen in configure vs make? (Idk about others, but I had no idea what happened during autogen vs configure vs make when I was starting out. And it's still a pretty black-box-ish part of my workflow. That could be a good question too.)
- Questions with instructions, e.g. "did you try building this way," or "did you try passing in this flag during ./configure," etc. This might help reward those who prepare ahead of time + help people out if they haven't prepared

0xB10C:
Danke für die Kommentare und vor allem für die zusätzlichen Fragen! I probably won't have time to add more questions before Monday. I think it could make sense to merge this now, so people preparing don't prepare for the wrong PR.

glozow:
Ich fixupte a tiny bit and mergete, feel free to open another PR with questions whenever convenient :) thanks @0xB10C!

PR #454, opened 2022-01-12
Show irc channel name more prominently

jnewbery:
We lost the channel name from the front page somewhere along the way.


jnewbery:
Hopefully people will be able to find the right channel now

https://www.erisian.com.au/bitcoin-core-dev/log-2022-01-12.html#l-306

glozow:
ACK

PR #470, opened 2022-02-02
Update libsecp #748

jnewbery:
- rename file to indicate that it's a libsecp PR, not a bitcoin core PR (no change to the generated site)
- link to bitcoin-core/libsecp instead of ElementsProject/secp256k1-zkp
- add site-wide aliases for the Bitcoin Core, BIPs and libsecp repos, and a single page variable for a recent commit in libsecp to avoid repetition and very long URLs in the markdown files


glozow (commented):
waow cool, very nice

PR #477, opened 2022-02-24
Add logs for n24098

stickies-v:


jnewbery:
ACK. Logs match mine. Thanks @stickies-v

PR #490, opened 2022-03-30
add logs for v23rc testing

glozow:


glozow:
@stickies-v good point. I'll correct the timestamps when I get a chance, sorry about that.

PR #505, opened 2022-05-12
Add preparatory resources to n24148

stickies-v:
Temporary update while finalizing full notes and questions.

### DIFF
--- a/_posts/2022-05-18-#24148.md
+++ b/_posts/2022-05-18-#24148.md
@@ -10,11 +10,20 @@ status: upcoming
 commit:
 ---
 
-_Notes and questions to follow soon!_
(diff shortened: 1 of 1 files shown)


glozow:
Thanks for getting started on notes/questions! Will you be adding them to this PR later or should I review it now?

stickies-v:
I was thinking to create a follow up PR later this weekend with fleshed out notes and questions as I haven't finished yet, but given the scope of the PR I figured having some prep materials would be a starting point already - for those interested.

glozow:
Ok sure, we'll put these up for people who want to get started now, and add the rest later. No rush. Thanks!

PR #507, opened 2022-05-17
n24148: fix bad commit message and add 2 questions

stickies-v:
This was meant to be force pushed as per darosior's comments in #506 but it looks like I [didn't actually commit](https://github.com/bitcoin-core-review-club/website/compare/8256a281662dcd0fead180ac0f2aa2965ed6b820..df89aad40fb433befcd73071c60d40f25e20bdc8) anything, fixed now.

### DIFF
--- a/_posts/2022-05-18-#24148.md
+++ b/_posts/2022-05-18-#24148.md
@@ -16,7 +16,7 @@ commit:
 
 This is a 2-part Review Club. Since we haven't covered Miniscript before, we first take some time to get familiar with the general concepts before diving into the actual PR.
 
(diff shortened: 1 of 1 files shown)


_posts/2022-05-18-#24148.md
```diff
@@ -54,6 +56,8 @@ This is a 2-part Review Club. Since we haven't covered Miniscript before, we fir
```
glozow, line 56: (a few lines up)

> How do we keep track of a `Node`'s type and type properties?

Is this a type-o, maybe missing a link for disambiguation, or am I missing something? 😅

stickies-v: I've added in hyperlinks to avoid ambiguity, does this clear it up?

PR #514, opened 2022-06-07
small fixups to n24148

glozow:
Mostly little nitpicks.

- Suggest "it may be helpful to review" instead of "it may be helpful to gloss over." It threw me off a little bit because I interpret "gloss over" to mean "handwave it away" or "pretend you covered something even though you didn't."

- Embed link instead of pasting as raw text

- Use semicolon instead of comma


stickies-v:
(unsolicited) ACK

glozow:
Thank you @stickies-v :)

PR #517, opened 2022-06-15
n24170: typo: paper from 2015 instead of 2005

stickies-v:


glozow:
merging since it seems @mzumsande has +1ed

mzumsande:
Yes, of course, forgot to ACK!

PR #542, opened 2022-08-11
add 25720 logs

dergoegge:


stickies-v (approved):
ACK - logs match with mine, just pushed some missing quotes in the questions. Thank you @dergoegge!

PR #159, opened 2020-03-20
Convert auto_logs_markup global plugin to callable tag

jonatack:
Now that build speed appears to no longer be an issue thanks to #154, this PR updates @harding's callable filter suggestion implemented by @jnewbery in https://github.com/bitcoin-core-review-club/bitcoin-core-review-club.github.io/tree/pr147.2

Co-authored-by: John Newbery john@johnnewbery.com
Co-authored-by: David A. Harding <dave@dtrt.org>

PR #585, opened 2022-11-09
Improve IRC nickname parsing

stickies-v:
Fixes the IRC nickname parsing issue reported in #584 and reverts the manual fix since it is no longer necessary.

The updated nickname regex now matches the data _after_ the time stamp (`IRC_NICK.match(line[TIME_SIZE_PLUS_1..-1]).to_s`) and requires the `IRC_NICK` pattern to be at the beginning (`^`) of that (shortened) string, while allowing additional leading whitespaces in a non-capturing group (`(?:\s*)`). The leading whitespaces tolerance is because [n24098](https://github.com/bitcoin-core-review-club/website/blob/master/_posts/2022-02-23-%2324098.md?plain=1) has double whitespaces between timestamp and nickname (guess who added those logs?!). Instead of just fixing the logs in n24098, I think it makes sense to not have the parsing be unnecessarily strict?

cc @LarryRuane for reporting initial issue and @jonatack for previously improving this logic in #159 (making you the de facto expert 🥇)

## Testing
To verify the impact of this PR, one approach is to build the website using both versions and compare the diff in the `_site` folder which contains the html files.
To build, see the instructions in [README.md](https://github.com/bitcoin-core-review-club/website/blob/b3dd96a22a18acd8a7bb95940886ae9eaf4e7da1/README.md)

A simple script that does all of this:
```sh
# Create site contents on unfixed master (5110706) branch and copy to `unfixed/` dir
git checkout 5110706 && \
rm -rf _site/ && rm -rf unfixed && mkdir unfixed &&\
make && \
cp -r _site/ unfixed/ && \

# Create site contents on fixed PR branch and copy to `fixed/` dir
git checkout 3d3fd24 && \
rm -rf _site/ && rm -rf fixed && mkdir fixed &&\
make && \
cp -r _site/ fixed/  && \

# Compare the differences
# remove `-q` to see detailed diff
diff -r -q unfixed/ fixed/
```

### DIFF
--- a/_plugins/auto_logs_markup.rb
+++ b/_plugins/auto_logs_markup.rb
@@ -17,7 +17,7 @@ module Jekyll
     HH_MM = /^([0-1][0-9]|[2][0-3]):[0-5][0-9] .*/
 
     # Regex to select IRC <nick>.
-    IRC_NICK = /<.+?>/
+    IRC_NICK = /^(?:\s*)(<.+?>)/
 
     # Regex to select "<" and ">" chars.
     LT_GT = /[<>]/
@@ -55,7 +55,7 @@ module Jekyll
         # Separate the log line into useful parts.
         lineno  = "#{NON_BREAKING_SPACE * (LINE_DIGITS - index.to_s.size)}#{index}"
         time    = line[0..TIME_SIZE]
-        name    = IRC_NICK.match(line).to_s
+        name    = IRC_NICK.match(line[TIME_SIZE_PLUS_1..-1]).to_s
         nick    = name.gsub(LT_GT, '').strip
         color   = colors[nick] || (color_index = (color_index + 1) % NUM_COLORS ;
                                    colors[nick] = COLORS[color_index])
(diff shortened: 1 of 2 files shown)


stickies-v:
Oh I forgot to add, this also (unintentionally) fixes an issue with n24098 where currently on https://bitcoincore.reviews/24098, an additional `>` bracket is added onto every line because of the double whitespace mentioned in OP:

image

When running a diff on the `_site/` folder before and after this fix, I'm only getting changes on these two posts (and feed.xml, but I don't think we care about that?):

```sh
colordiff -r -q unfixed/ fixed/
Files unfixed/_site/24098.html and fixed/_site/24098.html differ
Files unfixed/_site/26265.html and fixed/_site/26265.html differ
Files unfixed/_site/feed.xml and fixed/_site/feed.xml differ
```

LarryRuane:
concept ACK, I'm not set up to test this.

stickies-v:
Thanks @LarryRuane. FYI, I think empirically testing this PR is actually not that much work, I've edited the PR to include the testing steps I used.

LarryRuane:
Ok I'll test in the next few days, looks great!

glozow:
Thanks @stickies-v! Thanks for the testing script, my output is also

```
Files unfixed/24098.html and fixed/24098.html differ
Files unfixed/26265.html and fixed/26265.html differ
Files unfixed/feed.xml and fixed/feed.xml differ
Files unfixed/sitemap.xml and fixed/sitemap.xml differ
```

If @jonatack or someone who understands the parsing better than me wants to review that would be great. But LGTM so I'm happy to merge.

PR #592, opened 2022-11-18
add n26152 notes

glozow:
I may add more questions. This PR is quite large 🤔

### DIFF
--- a/_posts/2022-11-23-#26152.md
+++ b/_posts/2022-11-23-#26152.md
@@ -7,17 +7,165 @@ authors: [Xekyo]
 components: ["wallet"]
 host: glozow
 status: upcoming
-commit:
+commit: 898ad9d590
 ---
 
-_Notes and questions to follow soon!_
-
-<!-- TODO: Before meeting, add notes and questions
 ## Notes
 
+- The wallet *funds* a transaction by selecting inputs to cover its payment amount(s) and the fee
+  at the user's target feerate. This process is known as **coin selection**, which we have discussed in
+previous review clubs [#22009](/22009), [#17526](/17526) and [#17331](/17331).
+
+    - Notably, each candidate coin is considered using an "effective value," introduced in
+      [PR #17331](https://github.com/bitcoin/bitcoin/pull/17331). This deducts the cost to spend
+      this input at the target feerate from its `nValue`.
+
+- Since [PR #7600](https://github.com/bitcoin/bitcoin/pull/7600), the
+  [`BlockAssembler`](https://github.com/bitcoin/bitcoin/blob/aeb395dcdbfe2b1a6c77ff218939a18afde3add9/src/node/miner.h#L164)
+algorithm has used ancestor feerate, rather than individual feerate, to select transactions for
+inclusion in blocks. This strategy also enables users to fee-bump transactions using [Child Pays for
+Parent (CPFP)](https://bitcoinops.org/en/topics/cpfp/). We have discussed the `BlockAssembler`
+implementation in a [previous review club meeting](/24538).
+
+- On the flip side of CPFP, a transaction's "effective" feerate depends on the UTXOs used to fund
+  it: if unconfirmed UTXOs are used, since the transaction cannot be mined without its ancestors,
+its "effective" feerate may decrease. Even while not trying to boost an unconfirmed transaction
+using CPFP, if the wallet funds a transaction using unconfirmed UTXOs with feerates lower than the
+target feerate, it may underestimate the amount of fees to put on this transaction. This issue has
+been an open problem for years; see
+[#9645](https://github.com/bitcoin/bitcoin/issues/9645),
+[#9864](https://github.com/bitcoin/bitcoin/issues/9864), and
+[#15553](https://github.com/bitcoin/bitcoin/issues/15553).
+
+- Goal: when funding a transaction using unconfirmed inputs, in addition to funding the payment(s)
+  and this transaction's fee, also include the fees necessary to bump those unconfirmed transactions
+to the target feerate.
+
+- One naive solution would be to select all the inputs, and then calculate the cost of bumping
+  unconfirmed ancestors and add that to the fees. However, if the selected inputs cannot cover the
+cost of ancestor fees, the selection process needs to be repeated; this strategy would (re)introduce
+a looped coin selection algorithm and make changeless solutions much less common.
+
+- Consider another solution which updates the "effective value" of a coin, reducing it by the fee
+  necessary to bump its ancestor set to the target feerate: `effective_value = nValue -
+cost_to_spend - min(0, target_feerate * ancestor_size - ancestor_fees)`. For example, if the target
+feerate is 20sat/vB, the UTXO has type P2WPKH and value 100ksat, and transaction's ancestor fee and
+vsize are 1200sat and 600vB respectively, the coin's effective value would be 100000sat -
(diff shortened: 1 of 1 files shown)


glozow:
ccing @Xekyo if you're interested in taking a look

_posts/2022-11-23-#26152.md
```diff
@@ -7,17 +7,152 @@ authors: [Xekyo]
...
+- Since [PR #7600](https://github.com/bitcoin/bitcoin/pull/7600), the
+  [`BlockAssembler`](https://github.com/bitcoin/bitcoin/blob/aeb395dcdbfe2b1a6c77ff218939a18afde3add9/src/node/miner.h#L164)
+algorithm has used ancestor feerate, rather than individual feerate, to select transactions for
+inclusion in blocks. This strategy also enables users to fee-bump transactions using [Child Pays for
+Parent (CPFP)](https://bitcoinops.org/en/topics/cpfp/). We have discussed the `BlockAssembler`
+implementation in a [previous review club meeting](/24538).
+
+- On the flip side of CPFP, when creating a transaction at a target feerate, if the wallet selects
+  UTXOs from lower-feerate unconfirmed transactions, it may underestimate the amount of fees to put
+on this transaction. This issue has been an open problem for years; see
```
stickies-v, line 32: I've expanded it a bit, there's a lot to cover here.
```suggestion
- The effective feerate of a transaction depends on the confirmation status of all of its parent transactions. If all parents are already confirmed, the effective feerate equals the feerate of the transaction itself. However, each unconfirmed parent transaction increases/decreases the effective feerate if the parent feerate was higher/lower than the target feerate. This behaviour is exactly the same as what we require for CPFP, but with CPFP we focus on boosting the parent transaction - whereas this PR deals with the wallet ensuring the child transaction receives the effective feerate the user intends. This issue has been an open problem for years; see
```

_posts/2022-11-23-#26152.md
```diff
@@ -7,17 +7,152 @@ authors: [Xekyo]
...
+
+    - Multiple UTXOs may come from the same transaction or share the same ancestry. We don't need to
+      bump those transactions more than once.
+
+    - An ancestor's feerate may be higher than the target feerate.
+
+    - An ancestor may have a low feerate, but be already fee-bumped by another transaction in the
+      mempool.
+
+    - If the transaction may replace something in the mempool, one or multiple transactions may be
```
stickies-v, line 64: Can this affect ancestors? I'd think this would only evict descendants, and thus not affect ancestor feerate?
```suggestion
    - If the transaction replaces an existing transaction in the mempool, one or multiple transactions may be
```

_posts/2022-11-23-#26152.md
```diff
@@ -7,17 +7,152 @@ authors: [Xekyo]
...
 ## Questions
+
 1. Did you review the PR? [Concept ACK, approach ACK, tested ACK, or NACK](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#peer-review)?
--->
+
+1. What issue does this PR solve?
+
+1. In
+   [`CalculateCluster`](https://github.com/bitcoin-core-review-club/bitcoin/commit/995107782a1a512811d54f7abf29249f351a7cbf#diff-c065d4cd2398ad0dbcef393c5dfc53f465bf44723348892395fffd2fb3bac522R1218),
+what does a transaction's "cluster" include?
```
stickies-v, line 100:
```suggestion
what does a transaction's "cluster" consist of?
```

stickies-v (approved):
Wow, very thorough notes and question, thank you @glozow! Left a few phrasing suggestions but nothing blocking. I'm thinking if maybe this should be 2 clubs? A lot of concepts and questions to cover.

_posts/2022-11-23-#26152.md
glozow: Done

_posts/2022-11-23-#26152.md
glozow: I took a few issues with the wording but otherwise taken:

- the effective feerate depends on ancestors and all descendants of those ancestors, and not just parents
- a parent cannot increase the effective feerate of the transaction

_posts/2022-11-23-#26152.md
glozow: took the grammaf fix and added an example

_posts/2022-11-23-#26152.md
glozow: done

glozow:
Thanks @stickies-v!

> I'm thinking if maybe this should be 2 clubs? A lot of concepts and questions to cover.

Yeah I agree, let's see how far we get next Wednesday. If we have time and people are interested in doing another session, we can do this on the 30th as well? A break for you :D

_posts/2022-11-23-#26152.md
stickies-v: > the effective feerate depends on ancestors and all descendants of those ancestors, and not just parents

Whoops, yes I meant ancestors instead of parents - sorry for the confusion.

> a parent cannot increase the effective feerate of the transaction

Interesting, I hadn't considered that asymmetry (that it can only decrease) - but you're absolutely right.

_posts/2022-11-23-#26152.md
```diff
@@ -7,17 +7,165 @@ authors: [Xekyo]
...
+  the block assembly algorithm designed to calculate the cost to bump an unconfirmed UTXO to a
+target feerate.  It uses `MiniMiner` to include the fees to bump any unconfirmed ancestors when
+funding a transaction.
+
+- `MiniMiner` operates on a limited set of mempool entries rather than the entire mempool, and does not care
+about consensus rules such as block size and sigop limits. The limited set includes the "clusters"
+of each unconfirmed transaction from which the wallet might spend a UTXO from. A *cluster* includes
+all mempool entries "connected" to a transaction. This includes parents and children,
+ancestors and descendants, as well as "siblings" (a transaction that shares a parent) and
+"coparents," (a transaction that shares a child), etc.
```
Xekyo, line 82: It might help to point out that any ancestor and descendant of any transaction in the cluster is also in the cluster:

```suggestion
all mempool entries "connected" to a transaction. This includes direct parents and children,
ancestors and descendants, as well as transitively the ancestors and descendants of any connected transaction, e.g. "siblings" (a transaction that shares a parent), "coparents," (a transaction that shares a child), etc.
```

_posts/2022-11-23-#26152.md
```diff
@@ -7,17 +7,165 @@ authors: [Xekyo]
...
+
+- The PR adds a `CTxMemPool::CalculateCluster` function to calculate a set of transactions'
+  clusters. It uses the epoch-based traversal strategy described in
+[PR #17268](https://github.com/bitcoin/bitcoin/pull/17268) and first introduced in
+[PR #17925](https://github.com/bitcoin/bitcoin/pull/17925).
+
+- `MiniMiner` provides two functions: `CalculateBumpFees()` provides the bump fee for each UTXO
+  independently, ignoring any cases where there are shared ancestries. This function is intended for
```
Xekyo, line 90: This sounds ambiguous to me. How about:

```suggestion
  independently, as if it were spent alone (ignoring that another UTXO may share its ancestry). This function is intended for
```

Xekyo (commented):
Very cool! Thanks for the elaborate description. Got a couple suggestions.

PR #598, opened 2022-12-07
add n25574 logs

glozow:


stickies-v (approved):
LGTM but it looks like Netlify is down (tests failing without any logs showing) so let's merge tomorrow, hopefully they'll be back up by then? I assume if we merge now the deployment will fail too..

stickies-v:
Did an empty force-push to retrigger CI, looks like everything's fine now so merging this.

glozow:
Thanks @stickies-v!

PR #602, opened 2022-12-16
notes and questions for n26695

glozow:


stickies-v (approved):
LGTM - just pushed a small typo fix. Thanks for prepping and hosting!

I think you already have enough questions, but if you want more perhaps could be useful to talk about the complexity/scaling of this benchmark? E.g. why 1000 txns, impact of increasing/decreasing. But I'll merge this already, I think it's great as is.

glozow:
Thanks @stickies-v!

PR #612, opened 2023-01-23
add some hosting pointers

glozow:
Been trying to incorporate some of this informally, and makes sense to add to the docs.

(1) "What was your review approach?" to the templated first question seems to be helpful.

(2) Giving people more time to prepare is a common suggestion. Also attempt to be a bit better with scheduling since we seem to frequently run into weekends/Mondays.


glozow:
Yeah I think the only problem with hosting merged PRs is people don't feel as motivated to review them (it seems). fwiw I wish that wasn't the case, as the purpose here isn't to get PRs merged, but to learn... I look at old PRs all the time :shrug: maybe we should deliberately review some old ones?

PR #619, opened 2023-02-02
n26847: fix typo

mzumsande:
That was me trying to remember my shortcut for markdown preview :smile:


stickies-v (approved):
Oh wow that slipped through the review cracks a fair amount of times haha. Thanks for fixing!

PR #633, opened 2023-03-02
announce n27006

stickies-v:
The PR doesn't have any labels so I've manually added the ["refactoring", "resource usage"] which seemed most appropriate.

FYI @furszy.

### DIFF
--- a/_posts/2023-03-15-#27006.md
+++ b/_posts/2023-03-15-#27006.md
@@ -1,0 +1,28 @@
+---
+layout: pr
+date: 2023-03-15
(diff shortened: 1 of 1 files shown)


glozow:
Awesome, thanks @stickies-v!

furszy:
Cool stickies-v 👌🏼.

Just as an extra information, have been weighting a different approach to fulfill the same purpose.
Approach that is currently not possible due the non-deleted copy constructor of the `CBlockIndex` class but.. leaving that point aside, the move of `nFile` to atomic should also work (pending memory footprint check). The read race is benign, callers can handle the error as if the block would had never been there, independent of the returned file position.

PR #644, opened 2023-03-22
add post for april 5th

josibake:


stickies-v (approved):
Thanks for hosting @josibake!

- pushed a small title typo
- aligned components with the Bitcoin Core PR labels

You probably already know this, but you can automatically generate these posts with [`rake`](https://bitcoincore.reviews/hosting#:~:text=in%20the%20moment.-,Making%20a%20New%20Post,-Fork%20the%20website) in which case all the PR details are pulled automatically (but you may not want to install the dependencies). As the PR doesn't have any labels atm, the components would have been empty though, I think `descriptors` is the most relevant.

josibake:
> As the PR doesn't have any labels atm, the components would have been empty though, I think descriptors is the most relevant.


yeah, annoyingly, the post fails to build if labels are empty, so i added them manually after the fact. agree that descriptors is the best one.

we should probably handle the case where labels on the PR is empty more gracefully, maybe by prompting the user to manually add one?

stickies-v:
Interesting, I wasn't aware of that - would you mind opening a quick issue so we can keep track of it? Otherwise I'm happy to do it too. 

Also now noticing that rake automatically sanitizes PR titles by lowercasing any non-leading capitalized letters, so apologies for unnecessarily changing that - wasn't aware.

PR #645, opened 2023-03-22
print a warning when there are no labels from github

josibake:
probably not the most elegant way to do this, so happy for suggestions on how to improve it!

without this, the post will fail to build when running `make preview`, but it fails with a really obtuse error:

```
Configuration file: /home/josibake/website/_config.yml
            Source: /home/josibake/website
       Destination: /home/josibake/website/_site
 Incremental build: enabled
      Generating... 
       Jekyll Feed: Generating feed for posts
  Liquid Exception: Could not locate the included file 'ERROR_92_MISSING_TOPIC_CATEGORY' in any of ["/home/josibake/website/_includes"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source. in meetings-components.html
```

took me longer than id like to admit to troubleshoot it.

at least the warning will point people in the right direction.. even better would be a command line arg for passing a label

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -177,7 +177,7 @@ def create_post_file!(filename, response, date, host)
   title = parse_title(response['title'])
   components = parse_valid_components(response['labels'])
   if components.empty?
-    puts "No labels from Github, you will need to add one manually from #{COMPONENTS}"
+    puts "No label assigned to the PR yet; you will need to add one manually from #{COMPONENTS}"
   end
 
   puts "GitHub PR title:  \"#{response['title']}\""
--- a/Rakefile
+++ b/Rakefile
@@ -176,6 +176,9 @@ end
 def create_post_file!(filename, response, date, host)
   title = parse_title(response['title'])
   components = parse_valid_components(response['labels'])
+  if components.empty?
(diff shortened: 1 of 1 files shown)


josibake:
@stickies-v per my comment on https://github.com/bitcoin-core-review-club/website/pull/644#issuecomment-1479867380

glozow:
Thanks :) concept ACK

josibake:
> During the two years I was co-maintaining this repo, I would regularly check if it contained all the current labels in bitcoin/bitcoin, and add/update any that were missing.

ah, sorry for the confusion, I could have been a bit more clear in my description. This is not about a mismatch between what exists on GitHub vs what exists here, although now that you mention that, I think that would also cause the same problem! 

For example:

* GitHub PR has a shiny new label
* It doesn't exist in our components list here
* New post ends up without any labels
* Fails to build

I'm not a ruby guy and I think this is good enough for now to at least prompt the user to manually add a label (regardless of the reason for why its missing), but if someone is feeling extra motivated I suppose you could have it print separate errors for a) no labels on the github PR and b) labels on the PR but not in our components list

PR #647, opened 2023-03-23
Ask for user input when no components are fetched from GitHub

stickies-v:
Without components, the project cannot build. With this change, the user is queried to manually input components on the command line if none are found. The input is sanitized to ensure data stays in line with `COMPONENTS`.

Follow up on #645

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -173,12 +173,26 @@ rescue *HTTP_ERRORS => e
   "Error #{e.inspect}"
 end
 
+def get_nonempty_components(gh_labels)
+  # Parses the GitHub labels, and requires user input if no valid components were found
+  components = parse_valid_components(gh_labels)
+  if components.empty?
+    puts "No label assigned to the PR yet; you will need to add one or more (comma-separated) manually from #{COMPONENTS}"
+    while true
+      components_input = gets.gsub(/['"]/, '').split(',').map(&:strip).map(&:downcase).uniq
+      if (components_input - COMPONENTS).empty?
+        break
+      end
+      puts "Components #{components_input - COMPONENTS} are invalid, please try again"
+    end
+    components = components_input
+  end
+  return components
+end
+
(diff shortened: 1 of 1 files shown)


josibake:
Concept ACK

nice, thanks for the follow-up!

Rakefile
```diff
@@ -173,12 +173,26 @@ rescue *HTTP_ERRORS => e
...
-  title = parse_title(response['title'])
-  components = parse_valid_components(response['labels'])
+def get_nonempty_components(gh_labels)
+  # Parses the GitHub labels, and requires user input if no valid components were found
+  components = parse_valid_components(gh_labels)
   if components.empty?
-    puts "No label assigned to the PR yet; you will need to add one manually from #{COMPONENTS}"
+    puts "No label assigned to the PR yet; you will need to add one or more (comma-separated) manually from #{COMPONENTS}"
+    while true
+      components_input = gets.split(',').map(&:strip).map(&:downcase).uniq
```
josibake, line 182: The labels are printed out with double quotes, so I included double quotes when passing my labels, which caused it to not recognize the label. I think it would be better to strip quotes so that it doesn't matter if the user uses quotes or not:

```suggestion
      components_input = gets.gsub(/['"]/, '').split(',').map(&:strip).map(&:downcase).uniq
```

josibake (approved):
ACK https://github.com/bitcoin-core-review-club/website/pull/647/commits/059b79733a35ad2be7404e38797c46acdb89fcbe

I added a small suggestion which I think makes it a little more user friendly, but also fine as is

Rakefile
stickies-v: Makes sense, added your suggestion, thanks!

josibake:
reACK https://github.com/bitcoin-core-review-club/website/pull/647/commits/21e10284b65033a502603ef3ef83ca1e3242c71c

PR #648, opened 2023-03-24
Add RSS and Twitter favicons

### LINKED ISSUE #625: Add RSS feed for notes
Instead of having people rely on Twitter or announcements in the IRC channel, I think having an RSS feed would be nice?

The most naive approach is to push an update for any change to the `_posts/` dir, or perhaps more elegantly could also allow people to individually subscribe to announcements, questions/nodes, and logs updates.

Found e.g. https://github.com/jekyll/jekyll-feed which seems to also supports tagging.

stickies-v:
Partially addresses #625, implementing jnewbery's [suggestion](https://github.com/bitcoin-core-review-club/website/issues/625#issuecomment-1481089733) to include favicons for our Twitter and Atom feed on the site header.

My jekyll and frontend skills are minimal, so I'm very open to suggestions on implementing this in a cleaner way, although I think it's fairly maintainable as is. To edit the icons, one can just update `_data/social-media.yml`.

image

### DIFF
--- a/_data/social-media.yml
+++ b/_data/social-media.yml
@@ -1,0 +1,9 @@
+atom:
+  href: 'feed.xml'
+  title: 'Atom Feed'
+  fa-icon: 'fa-rss'
+
+twitter:
+  href: 'https://twitter.com/bitcoincoreprs'
+  title: 'Twitter'
+  fa-icon: 'fa-twitter-square'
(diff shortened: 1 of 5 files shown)


jnewbery:
Concept ACK. I think it looks fine on a monitor, but a bit messy on a narrow display like a phone screen:



image

stickies-v:
I've switched the order of the elements so the icons keep floating above the title on narrow screens. This looks better, right?

image

jnewbery:
Looks good to me. utACK b7497c0

PR #650, opened 2023-03-28
Remove empty test_rake_posts_new test

jnewbery:
This test just asserts true and exits. I think it was originally intended to test the `rake posts:new` function, but was never implemented (see #45)


stickies-v (approved):
ACK aca592a3393cbab10cfb1303a702d9ed2a2c1c29

Test wasn't doing anything, and the new instructions work well.

PR #660, opened 2023-04-12
add irc logs for 27255 part 2

josibake:


stickies-v (approved):
LGTM - thanks @josibake !

PR #661, opened 2023-04-12
Add Pieter Wuille addendum to 27255-2 meeting log

jonatack:
These comments by Pieter on the IRC channel after the meeting add context that seems useful and educational for readers.


stickies-v (approved):
Thanks, useful to add indeed. LGTM.

PR #662, opened 2023-04-13
add notes and questions for n27039

stickies-v:
cc @pinheadmz - feel free to leave any comments/suggestions if you like (but no pressure).

### DIFF
--- a/_posts/2023-04-19-#27039.md
+++ b/_posts/2023-04-19-#27039.md
@@ -7,20 +7,48 @@ authors: [pinheadmz]
 components: ["block storage"]
 host: stickies-v
 status: upcoming
-commit:
+commit: 93c70287a
 ---
 
-_Notes and questions to follow soon!_
-
-<!-- TODO: Before meeting, add notes and questions
 ## Notes
 
+- Recommended reading from earlier review clubs:
+  - [17994: Flush undo files after last block write](/17994) to learn more about how blocks are downloaded and stored
+  - [16981: Improve runtime performance of --reindex](/16981) and [24858: incorrect blk file size calculation during reindex results in recoverable blk file corruption](/24858) to learn more about reindexing
+
+- The `blocks/` directory contains three datasets:
+  - the raw append-only `blknnnnn.dat` files containing the serialized block data
+  - the raw append-only `revnnnnn.dat` files containing the [undo data](https://bitcoin.stackexchange.com/questions/117887/why-undo-files-are-needed-in-the-bitcoin-core),
+  - the `blocks/index/` block index data which is stored in a LevelDB database. The index helps to quickly find blocks without having to re-scan all the data files.
+    - At runtime, the block index is loaded into [`BlockManager::m_block_index`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.h#L152). However, we also maintain statistics about the block files in [`BlockManager::m_blockfile_info`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.h#L117).
+
+- Flushing is the process of bringing the data that we keep in-memory in sync with the data we store on disk. We want to flush regularly enough to ensure that when bitcoind is unexpectedly killed, we may have to redo some work (e.g. re-download a couple of blocks) but we don't end up in an unrecoverable state (that would e.g. require a completely new IBD). On the other hand, the (generally quite slow) disk I/O can have negative performance implications when we flush too often.
+
+- A reindex, amongst other things, recreates the block index from the block files on disk. The block files themselves don't need to be updated: blocks are already assumed to not be stored sequentially in or across block files anyway.
+
+- This PR skips flushing to disk blocks whose positions are already known. Skipping this means that users can write-protect the blk files on disk which may be useful for security or even safely sharing that data between multiple bitcoind instances.
+
+- Some of the functions that are relevant to the reindexing process behave differently than can perhaps be expected from their name:
+  - [`BlockManager::SaveBlockToDisk()`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.cpp#L820) does not actually write anything to disk when (as is the case for a reindex) the known block position is passed (as `dbp`), but we still need to call it to update the blockfile info stats
+  - [`BlockManager::FindBlockPos()`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.cpp#L611) is mostly used to find the position in a blockfile where a new block is going to be appended. However, it also updates the blockfile info and may also trigger the flushing of block and undo files, which is why we still call it even if we already know the block's position.
+  - this behaviour arguably could arguably be improved upon, as for example done in [this WIP by mzumsande](https://github.com/bitcoin/bitcoin/compare/master...mzumsande:bitcoin:202207_refactor_findblockpos)
 
 
 ## Questions
 
 1. Did you review the PR? [Concept ACK, approach ACK, tested ACK, or NACK](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#peer-review)? What was your review approach?
--->
+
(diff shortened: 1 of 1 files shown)


_posts/2023-04-19-#27039.md
```diff
@@ -7,20 +7,47 @@ authors: [pinheadmz]
...
+
+1. After this PR, can the whole `blocks/` directory be made read-only? Why (not)?
+
+1. In [`FindBlockPos()`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.cpp#L611), what is the relevance of `fKnown` to this PR, conceptually? How does the behaviour of `FindBlockPos()` change when `fKnown==true` instead of `false`?
+
+1. What happens if bitcoind is killed after when block(s) have been processed, but before `FlushBlockFile()` is called? Does this PR affect that behaviour?
+
+1. In the `blockmanager_flush_block_file` unit test, why are we [pausing](https://github.com/bitcoin-core-review-club/bitcoin/commit/470ef396b5498d8689802c359a216d5a3c4749a5#diff-d6d633592a40f5f3d8b03863e41547de8751b874c1d20f129a616b9dd719b999R170) before starting a new blockfile? Can you think of issues with this approach, or alternatives?
+
+1. In the `blockmanager_flush_block_file` unit test, why do we still [only have two blocks](https://github.com/bitcoin-core-review-club/bitcoin/blob/470ef396b5498d8689802c359a216d5a3c4749a5/src/test/blockmanager_tests.cpp#L153-L154) in the file after having written `block3` to it?
```
pinheadmz, line 48: We *attempt* to write block3 but we know it won't be written. As mzumsande points out in a comment, we would never overwrite a block with different data, I'm just using this bizarre edge case to enforce the test.

_posts/2023-04-19-#27039.md
```diff
@@ -7,20 +7,47 @@ authors: [pinheadmz]
...
 ## Questions
 
 1. Did you review the PR? [Concept ACK, approach ACK, tested ACK, or NACK](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#peer-review)? What was your review approach?
--->
+
+1. After this PR, can the whole `blocks/` directory be made read-only? Why (not)?
+
+1. In [`FindBlockPos()`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.cpp#L611), what is the relevance of `fKnown` to this PR, conceptually? How does the behaviour of `FindBlockPos()` change when `fKnown==true` instead of `false`?
+
+1. What happens if bitcoind is killed after when block(s) have been processed, but before `FlushBlockFile()` is called? Does this PR affect that behaviour?
```
pinheadmz, line 44: This is actually a good question! And actually made writing the test pretty tricky. The reason is because the platform may flush at any time, usually when we `fwrite()`. So it's kinda like `FlushBlockFile()` could be called `EnsureFlush()` like, do it now if it hasn't happened already.

pinheadmz (approved):
Excited for this! There's a lot of spaghetti code and non intuitive function names throughout I hope that doesn't frustrate reviewers. I left a few notes for you below, not that you need to change anything but just some FYIs

stickies-v:
Thanks for the comments @pinheadmz, that's definitely helpful! I'm still figuring out some of the answers to the questions myself here. And like you say - the existing code doesn't always make things as clear as they could be...

_posts/2023-04-19-#27039.md
```diff
@@ -7,20 +7,47 @@ authors: [pinheadmz]
...
+
+- Flushing is the process of bringing the data that we keep in-memory in sync with the data we store on disk. We want to flush regularly enough to ensure that we can fully recover when bitcoind is unexpectedly killed, but not too regularly that the (generally quite slow) disk I/O slows down performance too much.
+
+- A reindex, amongst other things, recreates the block index from the block files on disk. The block files themselves don't need to be updated: blocks are already assumed to not be stored sequentially in or across block files anyway.
+
+- This PR skips flushing to disk those blocks that already have a known position in the datastore. Skipping this means that users can write-protect the blk files on disk which may be useful for security or even safely sharing that data between multiple bitcoind instances.
```
glozow, line 28: "the datastore" is somewhat new here, perhaps ?
```suggestion
- This PR skips flushing to disk blocks whose positions are already known. Skipping this means that users can write-protect the blk files on disk which may be useful for security or even safely sharing that data between multiple bitcoind instances.
```

_posts/2023-04-19-#27039.md
glozow: Re: pinheadmz's comment, maybe this instead?
```suggestion
1. In the `blockmanager_flush_block_file` unit test, why do we still [only have two blocks](https://github.com/bitcoin-core-review-club/bitcoin/blob/470ef396b5498d8689802c359a216d5a3c4749a5/src/test/blockmanager_tests.cpp#L153-L154) in the file after we have called `SaveBlockToDisk(block3)`?
```

_posts/2023-04-19-#27039.md
```diff
@@ -7,20 +7,47 @@ authors: [pinheadmz]
...
+
+- Some of the functions that are relevant to the reindexing process behave differently than can perhaps be expected from their name:
+  - `SaveBlockToDisk()` does not actually write anything to disk when (as is the case for a reindex) the known block position is passed (as `dbp`), but we still need to call it to update the blockfile info stats
```
glozow, line 31: A description of what it *does* do would be helpful, though not 100% necessary. Links to the functions would be nice though!

glozow (commented):
This looks great, thanks @stickies-v! Just some nits if you're interested.

_posts/2023-04-19-#27039.md
```diff
@@ -7,20 +7,47 @@ authors: [pinheadmz]
...
+
+- The `blocks/` directory contains three datasets:
+  - the `blknnnnn.dat` and `revnnnnn.dat` files which are raw append-only files containing the serialized block data and [undo data](https://bitcoin.stackexchange.com/questions/117887/why-undo-files-are-needed-in-the-bitcoin-core), respectively
+  - the `blocks/index/` block index data which is stored in a LevelDB database. The index helps to quickly find blocks without having to re-scan all the data files.
+    - At runtime, the block index is loaded into [`BlockManager::m_block_index`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.h#L152). However, we also maintain statistics about the block files in [`BlockManager::m_blockfile_info`](https://github.com/bitcoin/bitcoin/blob/19764dc143281376ea08e954018479ed10405b72/src/node/blockstorage.h#L117).
+
+- Flushing is the process of bringing the data that we keep in-memory in sync with the data we store on disk. We want to flush regularly enough to ensure that we can fully recover when bitcoind is unexpectedly killed, but not too regularly that the (generally quite slow) disk I/O slows down performance too much.
```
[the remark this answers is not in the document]
stickies-v: With fully recover, I mean that we don't end up in a corrupted state from which we can't recover, such as described e.g. in https://bitcoin.stackexchange.com/a/116784/129640 . This is something that (I think) we always (try to) ensure is true.

So in the cases where bitcoind is killed unexpectedly before flushing, this should be recoverable, e.g. by re-reading some blockfiles to update the index, etc.

Do you think this phrasing is more clear?

> We want to flush regularly enough to ensure that when bitcoind is unexpectedly killed, we may have to redo some work (e.g. re-download a couple of blocks) but we don't end up in an unrecoverable state (that would e.g. require a completely new IBD). On the other hand, the (generally quite slow) disk I/O can have negative performance implications when we flush too often.

_posts/2023-04-19-#27039.md
stickies-v: I feel like that's covered in the last part? 

> but we still need to call it to update the blockfile info stats

stickies-v:
Thanks @glozow - I've addressed all (but [one](https://github.com/bitcoin-core-review-club/website/pull/662#discussion_r1166649687)) your comments.

_posts/2023-04-19-#27039.md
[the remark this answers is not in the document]
glozow: Ah I see, that is clearer, thank you!

glozow:
Thanks @stickies-v and @pinheadmz ! Looking forward to review club!

PR #682, opened 2023-05-23
post: fix host username typo

ismaelsadeeq:


stickies-v (approved):
Thanks for fixing this @ismaelsadeeq, I used your IRC instead of GitHub username.

PR #707, opened 2023-08-02
add n28122 meeting 1 logs

glozow:

PR #708, opened 2023-08-02
n28122: fix log formatting issues

stickies-v:
Not sure what went wrong in #707, but some lines got garbled. Fixed (just using irccloud export and the script in https://bitcoincore.reviews/hosting)


glozow:
Ah oops - thanks for fixing @stickies-v! I think it was a text wrapping issue.

PR #716, opened 2023-08-24
n28165 notes

glozow:
I did push a tag, but I'm using the links directly from the PR since it's been merged and links won't change.

PR #717, opened 2023-08-30
n28165: continue numbering across meetings

stickies-v:
Looks like hardcoding the numbers in #716 didn't help, adding a kramdown tag.


glozow:
Ah sorry I didn't notice, thanks! Though I don't really mind the numbers starting over tbh.

stickies-v:
Main reason I thought it would be better for when people re-read the logs afterwards and references to questions are being made. Other than that, no preference for me either.

PR #734, opened 2023-11-29
n28690: Updates for newest patch set

TheCharlatan:


glozow:
The tag in the mirror hasn't been updated but we could do it. @stickies-v?

stickies-v (approved):
LGTM, thanks for updating this @thecharlatan. Updated the tag and the header, and added a quick note in case people had already started reviewing the previous version of the notes.

PR #736, opened 2023-12-12
announce n28956

stickies-v:
@glozow n28956 has no labels yet, P2P and Consensus seem most appropriate (checked with @dergoegge) so I added them here - would you mind adding them on the bitcoin repo too pls?


glozow:
Thanks @stickies-v!

PR #737, opened 2023-12-21
n28956: add notes and questions

stickies-v:
cc @dergoegge

### DIFF
--- a/_posts/2024-01-03-#28956.md
+++ b/_posts/2024-01-03-#28956.md
@@ -7,20 +7,58 @@ authors: [dergoegge]
 components: ["p2p", "consensus"]
 host: stickies-v
 status: upcoming
-commit:
+commit: 87c577e9742d7154826c755a7fe320f34fd54c81
 ---
 
-_Notes and questions to follow soon!_
-
-<!-- TODO: Before meeting, add notes and questions
 ## Notes
 
+- Bitcoin relies on the concept of time for operations such as the difficulty adjustment and transaction- or script-level timelocks. Since time is relative and clocks are notoriously difficult to keep synchronized, it is impossible to define a source of truth that does not depend on synchronization and authorities such as NTP servers. In a decentralized system, we must accept that nodes can have a different but equally valid view of the current time.
+
+- The Bitcoin network comes to consensus on time by requiring miners to commit to a timestamp in the [block header](https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch09.asciidoc#block-header). To prevent miners from being able to use excessively deviating timestamps in their blocks, nodes [verify](https://github.com/bitcoin/bitcoin/blob/3a0f54dd2402d74a5ac4304b3ad09014cfb25edf/src/validation.cpp#L3803-L3810) that the timestamp is later than the `Median Time Past` (calculated on the previous 11 blocks), and earlier than 2 hours from the current time.
+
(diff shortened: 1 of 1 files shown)


stickies-v:
Since I presume most folks are on holidays atm, I'm going to merge this already, but I'm more than happy to open a new pull if people have suggested improvements!

_posts/2024-01-03-#28956.md
```diff
@@ -7,20 +7,58 @@ authors: [dergoegge]
...
 ## Questions
 
+### Concept
+
 1. Did you review the PR? [Concept ACK, approach ACK, tested ACK, or NACK](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#peer-review)? What was your review approach?
--->
+
+2. Is it necessary for block headers to have a timestamp? If so, why?
+
+3. What is the difference between Median Time Past (MTP) and network-adjusted time? Which of these are relevant to the PR?
```
glozow, line 34:
```suggestion
3. What is the difference between Median Time Past (MTP) and network-adjusted time? Which of these is relevant to the PR?
```

_posts/2024-01-03-#28956.md
```diff
@@ -7,20 +7,58 @@ authors: [dergoegge]
...
+
+- Prior to this PR, the current time was calculated based on the network-adjusted time. During the handshake, the peer node's current time is compared to the system's current time, and the time difference is [stored](https://github.com/bitcoin/bitcoin/blob/3a0f54dd2402d74a5ac4304b3ad09014cfb25edf/src/net_processing.cpp#L3556). The network-adjusted time is then calculated by adding the median of these offsets to the system's current time, unless it exceeds the `-maxtimeadjustment` value [in which case a warning is emitted](https://github.com/bitcoin/bitcoin/blob/3a0f54dd2402d74a5ac4304b3ad09014cfb25edf/src/timedata.cpp#L84-L100).
```
glozow, line 19: This could clarify what is peer-specific and how that becomes our view of the network.

```suggestion
- Prior to this PR, the current time was calculated based on the network-adjusted time. When connecting to a peer, the peer's current time is compared to the system's current time, and the time difference is [stored](https://github.com/bitcoin/bitcoin/blob/3a0f54dd2402d74a5ac4304b3ad09014cfb25edf/src/net_processing.cpp#L3556). The network-adjusted time is then calculated by adding the median of all peers' offsets to the system's current time, unless it exceeds the `-maxtimeadjustment` value [in which case a warning is emitted](https://github.com/bitcoin/bitcoin/blob/3a0f54dd2402d74a5ac4304b3ad09014cfb25edf/src/timedata.cpp#L84-L100).
```

glozow (commented):
Thanks @stickies-v! Sorry for not getting to this earlier. Looks great, just a couple nits that you can feel free to ignore. I'll post the tweet shortly.

_posts/2024-01-03-#28956.md
stickies-v: It could be both, so I think "are" is the correct form?

_posts/2024-01-03-#28956.md
stickies-v: Thanks, I'll include this when I update the notes after the PR is rebased!

stickies-v: Included in https://github.com/bitcoin-core-review-club/website/pull/738

PR #739, opened 2024-01-03
replace broken web client link

glozow:
Providing an easy way for people to join via web client is nice and less discouraging to people who don't normally use irc. The original kiwi irc link doesn't work anymore, but this new link seems to work.


glozow:
Thanks @sr-gi for pointing out the previous link was broken

stickies-v:
Thanks @sr-gi. LGTM. I was wondering if the recent spam we've been having was due to the web client link being so easily scrapable (and thus thinking about if we should remove it), but clearly that wasn't the issue then :-D

PR #751, opened 2024-03-06
add logs for n39

glozow:


stickies-v (approved):
LGTM - pushed `status` to `past`. Thank you!

PR #760, opened 2024-05-22
announce n30132

stickies-v:
Small PR in terms of LoC, but lots of topics to cover wrt node startup, indexing, and the bugfix.


glozow (commented):
Cool! Thanks @stickies-v!

PR #764, opened 2024-06-12
contrib: use user.ghtoken for github api requests

glozow:
Avoid 403 rate limit errors

### DIFF
--- a/contrib/new_post.py
+++ b/contrib/new_post.py
@@ -6,9 +6,9 @@ import datetime
 import glob
 import json
 import os
+import subprocess
 import sys
-import sys
-import urllib.request
+from urllib.request import Request, urlopen
 
 MIN_PYTHON = (3, 9)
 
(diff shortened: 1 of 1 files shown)


contrib/new_post.py
```diff
@@ -54,6 +54,15 @@ class PullRequest():
     labels: list[str]
     user: str
 
+def git_config_get(option, default=None):
+    '''
+    Get named configuration option from git repository.
+    '''
+    try:
+        return subprocess.check_output(['git','config','--get',option]).rstrip().decode('utf-8')
```
jnewbery, line 62: `check_output()` is the old high-level API: https://docs.python.org/3/library/subprocess.html#older-high-level-api , and most new code uses `subprocess.run()`. Any particular reason to go for `check_output()` here?

jnewbery (approved):
LGTM! ACK 88501ff34b06fc54814a7523787e83c08c21a5bf

contrib/new_post.py
glozow: I ripped it from another script. Agree `run` seems preferred, changed.

stickies-v:
I think this is buggy when `user.ghtoken` is not set:

```
% ./contrib/new_post.py 30239 tdb3 2024-11-06
Github returned error HTTP Error 401: Unauthorized
```

`subprocess.run()` does not throw `CalledProcessError` for non-zero exit codes, that's `subprocess.check_output()` behaviour. As a result, without `user.ghtoken` set, an empty-string `Authorization` token is being added, leading to a `401`.

Will open a pull to fix.

PR #769, opened 2024-08-29
add announcement for v28 rc testing

glozow:
The testing guide is still wip, but we can announce now to schedule the meeting.


glozow:
cc @rkrux - thank you!

PR #771, opened 2024-09-12
add logs for v28rc testing

stickies-v:
Note: I've add the discussion post-#endmeeting too since it seems relevant.

Thanks again for hosting this @rkrux!


glozow:
Thank you both!

PR #787, opened 2025-02-03
clarifications for n31363

glozow:


ismaelsadeeq (commented):
nitty-nit in
Much of the work is done lazily. For example, calling `RemoveTransaction` does not actually mean its etry is deleted

s/its etry/its entry/g

PR #804, opened 2025-03-31
Add Docker support and instructions

stickies-v:
Adding Docker support makes it much easier for people to preview the website when making changes.

From the updated README:

```
You'll need [Docker](https://docs.docker.com/get-docker/) and [Docker
Compose](https://docs.docker.com/compose/install/) installed.

* Clone the repository and go into the directory
* Run `docker compose up` to start the development server with live
  reload
* Go to http://localhost:4000

```

(Note: to test this PR, it might be wise to add `--build` to the command, in case the Dockerfile changes: `docker compose up --build`)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+.git
+.gitignore
+.DS_Store
+_site
+.sass-cache
+.jekyll-cache
+.jekyll-metadata
+vendor
+.bundle
+Dockerfile
+docker-compose.yml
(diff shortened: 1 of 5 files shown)


glozow (approved):
ooh, I'm a fan. ACK

stickies-v:
Force-pushed to remove remaining documentation for a `test` service that I ended up removing because it was adding more scope creep than expected (may follow-up with future PR), RFM now.

PR #784, opened 2025-01-05
Fix social media icons in dark mode & upgrade ruby including dependen…

philippnagel:
closes #678

PR #805, opened 2025-03-31
chore: ruby upgrade

philippnagel:
* Carve out the changes made in #784 
* Update Ruby to 3.4.0
* Update deps and added deps that are not part of the standard library anymore


stickies-v:
Thanks for doing the upgrade too! To keep each PR focused, it would be best to have this branch be just for Ruby 3.4 upgrades, and keep #784 just for the dark mode fixes. To that end, I would:
- Drop the first 3 commits (except for Ruby upgrades)
- Update Dockerfile to use the `ruby:3.4-slim` image (you can then also easily test the website with `docker compose up --build preview`) and remove the specific `bundler` installation version (pinned at 2.4.22 because we were using such an old ruby version)
- Update README.md to mentioned Ruby 3.4 instead of 2.6

Ideally, we would also bump Jekyll to 4.4, but that's introducing a bunch of deprecation warnings, so if you're not very familiar with Jekyll (specifically, it was the Sass module) then we can leave that for another PR. Jekyll 3.x seems like it's still somewhat maintained, with 3.10.0 being [released](https://jekyllrb.com/docs/history/#v3-10-0) in 2024.

Gemfile
```diff
@@ -20,3 +20,9 @@ gem 'color_pound_spec_reporter'
 
 # HTTP bindings to libcurl - https://github.com/taf2/curb
 gem 'curb'
+
+gem "logger", "~> 1.7"
+
+gem "base64", "~> 0.2.0"
+
+gem "bigdecimal", "~> 3.1"
```
stickies-v, line 28: nit: these are Jekyll dependencies, so would be good to add them there. Some other dependencies can be updated to the most recent stable version too:

```
# Jekyll
gem "base64", "~> 0.2"
gem "bigdecimal", "~> 3.1"
gem "jekyll", "~> 3.10"
gem "jekyll-paginate", "~> 1.1"
gem "jekyll-sitemap", "~> 1.4"
gem "jekyll-feed", "~> 0.17"
gem "jekyll-seo-tag", "~> 2.8"
gem "kramdown-parser-gfm"
gem "logger", "~> 1.7"
```

PR #806, opened 2025-04-01
Fix XML feed by removing illegal characters and updating sanitization logic

### LINKED ISSUE #803: XML Parsing Error in feed.xml (Line 2056, Column 424)
I'm getting an error when I click on the feed icon.

SCREENSHOT-
Image

stickies-v:
The #771 logs added GS (Group Separator, 0x1D) and SI (Shift In, 0x0F) characters, likely because of text formatting in a rich-text IRC client. These are not valid XML characters, so they broke our XML feed (feed.xml)

Fix this by removing the illegal characters and update the instructions in `hosting.md` to specify the allowed range of characters as tab (\011), newline (\012), carriage return (\015), ASCII printable characters (\040-\176), and extended
ASCII/UTF-8 continuation bytes (\200-\377).

Testing the new `LC_ALL=C tr -dc '\011\012\015\040-\176\200-\377'` command on `feed.xml` did indeed only remove the illegal characters from #771, so I think this is a safe addition to our sanitization logic.

Fixes #803

### DIFF
--- a/_posts/2024-09-11-v28-rc-testing.md
+++ b/_posts/2024-09-11-v28-rc-testing.md
@@ -72,7 +72,7 @@ commit:
 17:20 <Murch[m]> Sorry, I don’t think my node is accessible from outside our network
 17:20 <glozow> mine too 😅 
 17:20 <Murch[m]> Happy to send some testnet4 coins to someone, though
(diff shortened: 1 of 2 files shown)


glozow (commented):
ACK, reproduced the issue and checked the fix. Thanks!

PR #807, opened 2025-04-01
chore: upgrade ruby, jekyll and deps.

philippnagel:


.ruby-version
```diff
@@ -1 +1 @@
-2.6.6
+3.4.0
```
stickies-v, line 1: Would be good to keep the newline. Also, I think patch version can be dropped and just keep "3.4".

Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /app
 
 COPY Gemfile* ./
 
-RUN gem install bundler -v 2.4.22
+RUN gem install bundler -v 2.6.2
```
stickies-v, line 13: Version can just be dropped here

```suggestion
RUN gem install bundler
```

Gemfile
```diff
@@ -3,12 +3,17 @@
...
+gem "jekyll-paginate"
+gem "jekyll-sitemap"
+gem "jekyll-feed"
+gem "jekyll-seo-tag"
 gem "kramdown-parser-gfm"
+gem "logger"
+gem "base64"
+gem "bigdecimal"
+gem "csv"
+gem 'webrick'
```
stickies-v, line 16: Any reason you didn't take the versions specified [here](https://github.com/bitcoin-core-review-club/website/pull/805#discussion_r2022633606)?

stickies-v, line 16: Why do we need "csv" and "webrick" gems?

Gemfile.lock
```diff
@@ -1,98 +0,0 @@
-GEM
```
stickies-v, line 1: What's the rationale behind deleting this file?

README.md
```diff
@@ -17,7 +17,7 @@ Compose](https://docs.docker.com/compose/install/) installed.
 ### Traditional Setup
 
 You'll need [Ruby and Jekyll](https://jekyllrb.com/docs/installation/) to run
-the site locally. Ruby 2.6 is recommended. Once they are set up:
+the site locally. Ruby 3.4.0 is recommended. Once they are set up:
```
stickies-v, line 20:
```suggestion
the site locally. Ruby 3.4 is recommended. Once they are set up:
```

.ruby-version
philippnagel: ACK

Dockerfile
philippnagel: ACK

Gemfile
philippnagel: Ruby is smart enough to determine the correct versions for our pinned version of Jekyll.

philippnagel: These were also removed from the core Jekyll file and had to be separately installed. The builds fail without these dependencies.

Gemfile.lock
philippnagel: The gemfile pins specific versions of the dependencies similar to e.g. how Node.js does it with package.lock. I removed this initially to fix the build errors. The CI seemed to fallback to an older version of Ruby. This was ultimately resolved by re-adding the specific .ruby-version file. Will commit the gemfile.lock again.

Gemfile
stickies-v: But jekyll should be 3.10, not 3.9. And pinning versions makes the build more deterministic. I think it'd be better to keep that as is but just bump to the most recent stable.

.ruby-version
philippnagel: Seems like the version needs to be explicitly defined.

Gemfile
stickies-v: They shouldn't be necessary with 3.10.

.ruby-version
stickies-v: Okay, thanks for trying it out. Nit: but newline is still removed.

Gemfile
philippnagel: Thanks for the heads-up. For some reason bundler decided that 3.9.5 was the most recent version. Since we are now commiting the lockfile again, the CI will always reference the versions there anyways. Do you still think it's better to pin the versions in the gemfile as well?

Gemfile.lock
```diff
@@ -35,64 +38,71 @@ GEM
...
     sass (3.7.4)
       sass-listen (~> 4.0.0)
     sass-listen (4.0.0)
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
+    webrick (1.9.1)
 
 PLATFORMS
-  ruby
+  x86_64-linux-gnu
```
stickies-v, line 87: This seems like it shouldn't be changed.

philippnagel: The lockfile is automatically generated from running the docker-compose command.

Gemfile
stickies-v: Yeah, specifying versions in Gemfile seems best practice, so I'd prefer keeping it that way.

philippnagel: Done in e109ea9

Gemfile.lock
stickies-v: I know, but it shouldn't be like this. I think it's because earlier changes you made introduced the `ffi (1.17.1-x86_64-linux-gnu)` line, which shouldn't be there..

This should fix it:

```
bundle lock --add-platform ruby
bundle lock --remove-platform x86_64-linux-gnu
```

philippnagel: Agree, seems like I overwrote this when running the preview natively outside the docker container. Fixed in 48c859b

stickies-v (approved):
LGTM! Can you please squash your commits, then I think it's ready for merge.

PR #809, opened 2025-04-02
feat: reactive dark mode.

### LINKED ISSUE #678: dark mode
hi @glozow 

would you accept a PR that enables dark mode for users whose operating system indicates [prefers-color-scheme:dark](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme#syntax) ? 

i'd be happy to take a look at that

philippnagel:
Closes #678

### DIFF
--- a/_includes/social-media-links.html
+++ b/_includes/social-media-links.html
@@ -3,7 +3,7 @@
     {% assign sm = site.data.social-media %}
     {% for entry in sm %}
         {% assign key = entry | first %}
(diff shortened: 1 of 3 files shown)


stickies-v:
I don't really like the approach taken here:
1. It seems a quite a bit of the newly introduced code in `default.html` doesn't actually do anything/is hallucinated (e.g. `applyColorMode` just sets a class element that does not exist/is not used)
2. there is no way to unset the local storage (except for going in system settings), so the user cannot revert back to automatic mode making this UX not great
3. I don't see the benefit of having a manual toggle when almost all browsers support `prefers-color-scheme`

I think #810 is much simpler and has my preference.

glozow:
Closing since #810 is merged

PR #810, opened 2025-04-03
style: automatically set dark mode, and fix invisible icons

### LINKED ISSUE #678: dark mode

stickies-v:
According to https://caniuse.com/?search=prefers-color-scheme, >96% of users have a browser that supports prefers-color-scheme.

Simplify UX, clean up code by following the browser/system dark mode settings and remove the manual toggle.

Also fixes #678 by rendering icons as white in dark mode.

Tested on Chrome and Firefox.

Alternative to #809

### DIFF
--- a/_includes/social-media-links.html
+++ b/_includes/social-media-links.html
@@ -3,7 +3,7 @@
     {% assign sm = site.data.social-media %}
     {% for entry in sm %}
         {% assign key = entry | first %}
-        <a href="{{ sm[key].href }}" title="{{ sm[key].title }}"><i class="fa {{ sm[key].fa-icon }}" style="color: #000000;"=></i></a>
+        <a href="{{ sm[key].href }}" title="{{ sm[key].title }}"><i class="fa {{ sm[key].fa-icon }}"></i></a>
     {% endfor %}
 </div>
 {% endif %}
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -5,9 +5,6 @@
 
 <body>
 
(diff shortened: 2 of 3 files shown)


glozow (commented):
seems to make sense, chat gpt assured me this is ok, yolo merging

polespinasa:
> yolo merging

:(
I think this was a mistake (personal opinion). I love dark mode, but the actual web page is too dark. I would rather use light mode for it, and this deleted the option to switch it to light mode.

IMO taking the default value from the system is nice, but the switch option should still there you don't want always the same color scheme for everything.

stickies-v:
Thanks for your feedback @polespinasa 

> but the actual web page is too dark

Perhaps you could open a PR that configures the dark mode values to ones that make more sense for you, and then we can evaluate if it makes sense to merge that for everyone?

polespinasa:
> PR that configures the dark mode values to ones that make more sense for you

Will do!


But still, I'm concept nack to deleting manual toggle. I don't think having that manual toggle is a bad UX experience, quite the opposite! [Here](https://ellemouton.com/posts/sphinx/) a great example of it (switch is at top-left).

polespinasa:
@stickies-v see https://github.com/bitcoin-core-review-club/website/pull/813

PR #813, opened 2025-04-10
Use more eye-friendly colors for the dark theme

polespinasa:
Change colors to more eye friendly as now the contrast is too heavy.
More info in this https://github.com/bitcoin-core-review-club/website/pull/810#issuecomment-2790082860

### DIFF
--- a/assets/css/all.sass
+++ b/assets/css/all.sass
@@ -46,6 +46,9 @@ body
 
 // Dark mode using media query
 @media (prefers-color-scheme: dark)
+  html, body
+    background-color: #1D1E20
+
   .site-container
     --bg-color: #1D1E20
     --text-color: #C4C4C5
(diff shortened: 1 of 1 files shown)


assets/css/all.sass
```diff
@@ -47,12 +47,12 @@ body
 // Dark mode using media query
 @media (prefers-color-scheme: dark)
   .site-container
```
stickies-v, line 49: If you add this, the background will also be properly coloured when the user scrolls past the end/beginning of a webpage:

```suggestion
  html, body
    background-color: #1D1E20

  .site-container
```

stickies-v (approved):
LGTM, this is much easier on the eyes indeed - thanks for suggesting this! Added one suggestion in comment, and also discovered that the `/hosting` page has a slight rendering issue with ```diff` code blocks, but that seems non-trivial to fix and is quite sporadic, so happy to leave that as-is (unless you have ideas):

image

PR #814, opened 2025-04-14
announce n31444 and n31981

glozow:


glozow:
cc @ismaelsadeeq and @ryanofsky  - thank you for hosting!

PR #825, opened 2025-05-07
add logs for n31375

stickies-v:


stickies-v:
FYI @ryanofsky I've highlighted this review club for the [latest optech newsletter](https://github.com/bitcoinops/bitcoinops.github.io/pull/2307/) (https://github.com/bitcoinops/bitcoinops.github.io/pull/2307/commits/e47f3f826d1ead69102fd68f185cd5aaf732c890)

ryanofsky:
Thanks and nice description in that commit! Only thing I might change is

> If and when it does in the future, the exemption in `security-check.py` should be removed and the calls fortified so they don't break the builds."

to

> If it does in the future, the exemption in `security-check.py` can be removed.

Since the foritify options are already enabled and the build shouldn't break if calls are added. it's just that security check doesn't have a good way of knowing fortify options are enabled so it needs this exception before they are added.

stickies-v:
Oh, thank you! I (mis)understood that from [here](https://bitcoincore.reviews/31375#l-121):

> If we add new calls they should be fortified based on build options.

I've pushed your fix in https://github.com/bitcoinops/bitcoinops.github.io/pull/2307/commits/687033b58ac8eda27f2fcd809b1cef8d283b1530.

PR #827, opened 2025-05-09
add notes and question for n32317

stickies-v:
I'm going to add some more meat to the notes, and maybe add a question or two later on, but I think this is already enough to online for now?

cc @TheCharlatan

### DIFF
--- a/_posts/2025-05-14-#32317.md
+++ b/_posts/2025-05-14-#32317.md
@@ -7,32 +7,68 @@ authors: [TheCharlatan]
 components: ["validation"]
 host: stickies-v
 status: upcoming
-commit:
+commit: 76a8f22c5c5cf48a9c36cc40db9224f0454917d0
 ---
 
-_Notes and questions to follow soon!_
-
-<!-- TODO: Before meeting, add notes and questions
 ## Notes
 
+### Motivation
 
+The bitcoinkernel project carves out validation logic into a separate, stateful library, allowing it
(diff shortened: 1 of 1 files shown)


_posts/2025-05-14-#32317.md
```diff
@@ -7,32 +7,68 @@ authors: [TheCharlatan]
...
+
+8. Why does [commit
+   `a7e4132`](https://github.com/bitcoin-core-review-club/bitcoin/commit/a7e41326234d3a381fdde0924af74c6561b10798)
+   use explicit template instantiation for `GetP2SHSigOpCount()`?
+
+9. Following on the previous question, why is there explicit instantiation of the
+   `<std::span<std::reference_wrapper<const Coin>>>` template when a `std::reference_wrapper<T>`
+   already [implicitly
+   converts](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper#:~:text=but%20they%20are,T%26)
+   to `T&`?
```
TheCharlatan, line 65: I actually need to dig into this reason again, but msvc was giving me a lot of pain when I tried to implicitly convert it.

stickies-v: Hmm, the answer I had in mind was that implicit conversion does not work for container value-types, because the reference wrappers are different objects allocated on the heap? (Although I think you might be thinking about the `GetCoin()` helper, which does not involve a container, and indeed confused me (and I have as an outstanding review comment) as to why we need it).

PR #844, opened 2025-09-28
move 33300 to Oct 8

marcofleon:
🙂


stickies-v (approved):
LGTM, better to not rush it in last minute.

PR #846, opened 2025-10-04
Fix commit links in 33300 notes

stringintech:
Commit links did not work because of the force push. Replaced them with the new link format (that should work regardless of force pushes) and new hashes.


stickies-v (commented):
Oh thank you for catching this @stringintech! For non-master commits we always link to our own bitcoin repo, because we keep tags for each version to keep links alive in case of force pushes, so I've pushed a commit to incorporate that. See [`hosting`](https://bitcoincore.reviews/hosting#:~:text=Push%20a%20tag%20of%20the%20branch) for more info.

@marcofleon are the notes still correct with these new (post-force-push) commits or do any further changes need to be made?

